PR #3, opened 2016-12-17
[review] bootstrap the review client

oliviertassinari:


src/review/webpack/browserBaseConfig.js
```diff
@@ -0,0 +1,62 @@
...
+      },
+      {
+        test: /\.woff$/,
+        use: 'url-loader?limit=100000',
+      },
+      {
+        test: /\.(jpe?g|png|gif|svg)$/,
+        use: 'file-loader', // Hash name by default
+      },
+      {
```
neoziro: No need to optimize images, we can do it before commit.

oliviertassinari: J'avoue, je vais supprimer ça.

PR #12, opened 2016-12-31
[server] Try having something live

oliviertassinari:
🌍


oliviertassinari:
We did it, it's live http://www.argos-ci.com/.

PR #27, opened 2017-01-07
[core] Add redux

oliviertassinari:
Also started configuring Google Analytics.


neoziro:
@oliviertassinari you forgot to set `sendHitTask` to `null` in development mode: https://developers.google.com/analytics/devguides/collection/analyticsjs/debugging

oliviertassinari:
@neoziro Wouldn't it be better to actually send data to a *staging* account?

Issue #25, opened 2017-01-03
Diff UI: review list


oliviertassinari:
ETQU, I can see a list of screenshot diffs for a single build.

### Motivation

That's the whole point of this project.

### Detail

The interface could be using a [Grid list](https://material.io/guidelines/components/grid-lists.html#). Something close to [Google Photo](https://photos.google.com/) should do it.

oliviertassinari:
I'm closing it, the interface is cheap but working. That should be enough to begin with.
#42.

PR #28, opened 2017-01-07
[review] Use RxJS to make the request

oliviertassinari:
Step after step, I'm setting up all the layers of the stack to display the screenshots 🎉 

Next step: implement the server graphQL resolver for that query. Data will use the seeds to begin with.
Next step: use the seeds data to actually render something with React.

Related to #25

### DIFF
--- a/src/review/routes/build/Build.js
+++ b/src/review/routes/build/Build.js
@@ -9,7 +9,6 @@ import ScrollView from 'modules/components/ScrollView';
 import LayoutBody from 'modules/components/LayoutBody';
 import BuildSummary from 'review/routes/build/Summary';
 import BuildScreenshots from 'review/routes/build/Screenshots';
-import graphQLClient from 'modules/graphQL/client';
 
 class Build extends Component {
   static propTypes = {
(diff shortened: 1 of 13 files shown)


src/review/routes/build/reducer.js
```diff
@@ -1,14 +1,19 @@
...
+import actionTypes from 'redux/actionTypes';
 
-function reducer(state, action) {
+function buildReducer(state, action) {
   if (state === undefined) {
     state = {};
   }
 
   switch (action.type) {
+    case actionTypes.BUILD_FETCH_TASK:
```
neoziro: You should only do this if we are in success isn't it?

src/review/routes/build/Build.js
```diff
@@ -54,4 +45,6 @@ class Build extends Component {
   }
 }
 
-export default connect()(Build);
+export default connect(state => ({
+  state,
```
neoziro: We don't need to connect all the state here but only the payload. Am I wrong?

oliviertassinari: You are right, that need some refinement.

src/review/routes/build/reducer.js
oliviertassinari: I'm planning adding a `watchTask` HoC, so having the progress state is going to help me out.

PR #29, opened 2017-01-08
feat(analytics): use a different tracking id in production

oliviertassinari:
- `UA-89989315-1`: production
- `UA-89989315-2`: development


neoziro:
Ca sert à quelque chose d'envoyer les données en dev?

neoziro:
@oliviertassinari is it useful to send data in development?

package.json
```diff
@@ -5,9 +5,9 @@
   "main": "index.js",
   "private": true,
   "scripts": {
-    "server:development": "nodemon src/server/bootstrap/bootstrap",
+    "server:development": "NODE_ENV=development nodemon src/server/bootstrap/bootstrap",
```
neoziro: Useless, default env will be development, we must get environment only from `config.get('env')`.

oliviertassinari:
The idea is to be closer to the production with a full end-to-end approach.
Not sending data mean we can't test stuff on the Google Analytics interface.

PR #34, opened 2017-01-14
feat(jobs): create a service to run jobs

neoziro:


.gitignore
```diff
@@ -2,3 +2,4 @@ node_modules
 /lib
 /server/static/
 *.log
+!.env
```
oliviertassinari: Why de we need to enforce `.env` files from not being ignored?

neoziro: I have it in my global .gitignore, it's always a good practice to ignore it I think.

PR #38, opened 2017-01-14
feat(cli): add command line interface package

### LINKED ISSUE #35: CLI: Create a command line interface for argos
In a separated repository, create a command line interface named "argos-cli".

When I type "argos" or "argos --help", I can see the help of the CLI.

neoziro:
Close #35


packages/argos-cli/src/index.js
```diff
@@ -0,0 +1,7 @@
+import program from 'commander';
```
oliviertassinari: 👍  for using commander

packages/argos-cli/src/index.js
```diff
@@ -0,0 +1,7 @@
...
+
+program.parse(process.argv);
```
oliviertassinari: ?
```jsx
if (!process.argv.slice(2).length) {
  program.outputHelp();
} else {
  program.parse(process.argv);
}
```

neoziro: I don't know if it works, but right now it is.

oliviertassinari: I'm asking because it's what we are doing the test cli for material. Avoiding running command when not argument are provided.

oliviertassinari:
For future inspiration https://github.com/callemall/material-ui/blob/next/test/cli.js.

PR #41, opened 2017-01-15
feat(api): add get buckets API

### LINKED ISSUE #36: API: get buckets
### Motivation

Used internally by the CLI.

### Detail

- Method: GET
- Endpoint: /buckets?token&branch
- Response body:
```graphql
[
  {
    id
    name
    commit
    branch
    jobStatus
  }
]
```

neoziro:
Close #36

### DIFF
--- a/src/server/routes/api.js
+++ b/src/server/routes/api.js
@@ -63,4 +63,16 @@ router.post('/buckets', upload.array('screenshots[]', 50), errorChecking(
   },
 ));
 
+router.get('/buckets', errorChecking(
+  async function (req, res) {
+    let query = ScreenshotBucket.query();
+
+    if (req.query.branch) {
+      query = query.where({ branch: req.query.branch });
+    }
+
+    res.send(await query);
+  },
+));
+
 export default router;
--- a/src/server/routes/api.test.js
+++ b/src/server/routes/api.test.js
@@ -2,13 +2,14 @@
 import path from 'path';
 import request from 'supertest';
 import { useDatabase } from 'server/testUtils';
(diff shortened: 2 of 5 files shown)


src/server/routes/api.test.js
```diff
@@ -27,4 +28,48 @@ describe('api routes', () => {
...
+      });
+    });
+
+    it('should returns buckets', () => {
+      return request(app)
+        .get('/buckets')
+        .set('Host', 'api.argos-ci.dev')
+        .expect(200)
+        .expect((res) => {
+          expect(res.body[0].name).toBe('test-bucket');
```
oliviertassinari: Whould a deep equal be better?

src/server/routes/api.js
```diff
@@ -63,4 +63,16 @@ router.post('/buckets', upload.array('screenshots[]', 50), errorChecking(
   },
 ));
 
+router.get('/buckets', errorChecking(
```
oliviertassinari: We could have exposed a graphql query. We might need to do the same for the front-end.

oliviertassinari:
Flaky flaky flaky 💥

src/server/routes/api.test.js
neoziro: No, I don't want to test the id.

src/server/routes/api.js
neoziro: I think it's not good to mix graphQL and REST, for the CLI I think it's better to be exclusively on REST for now. And I think it's better to separate endpoints.

src/server/routes/api.test.js
oliviertassinari: include then?

PR #48, opened 2017-01-22
chore(js): upgrade the dependencies

oliviertassinari:


src/server/routes/api.js
```diff
@@ -45,14 +47,18 @@ router.post('/buckets', upload.array('screenshots[]', 50), errorChecking(
           jobStatus: 'pending',
         })
 
+      const inserts = []
       for (const file of req.files) {
```
neoziro: map?

oliviertassinari: Ouais, j'avoue

PR #78, opened 2017-02-11
feat(review): hydrate the profile page

oliviertassinari:
I start wondering if redux, redux-rxjs is actually a good idea 😬 .
capture d ecran 2017-02-11 a 19 10 47


src/review/routes/profile/Profile.js
```diff
@@ -1,46 +1,97 @@
...
-                </Link>
-              </Text>
-            </li>
-          </ul>
-        </LayoutBody>
-      </ScrollView>
-    </ViewContainer>
-  )
-}
+  componentDidMount() {
```
neoziro: Not a good idea to mix view concern with data fetching concern if we want to switch from redux to another solution.

oliviertassinari: I agree, will use https://github.com/neoziro/recompact/tree/master/docs#lifecyclespec

PR #82, opened 2017-02-13
chore(npm): upgrade dependencies

oliviertassinari:
I have noticed one issue. After the synchronization, my email is gone 😆 .
capture d ecran 2017-02-14 a 00 06 17


src/server/models/User.js
```diff
@@ -15,7 +15,7 @@ export default class User extends BaseModel {
       githubId: { type: 'number' },
       accessToken: { type: 'string' },
       name: { type: 'string' },
-      email: { type: 'string' },
+      email: { type: ['string', null] },
```
oliviertassinari: Without this fix, the task is in an endless loop mode.

PR #91, opened 2017-02-19
feat: upload diff to s3

### LINKED ISSUE #87: Upload the diff to S3 in the job

neoziro:
Closes #87


neoziro:
@oliviertassinari yes but not very relevant.. We check that the field is not undefined and it was null.

Issue #86, opened 2017-02-16
Notify PRs on github


oliviertassinari:
- When the build is in progress
- When the build is finished
- When a user use the toggle validation button in the interface.

oliviertassinari:
Done.

PR #92, opened 2017-02-19
feat: notify GitHub

neoziro:
See #86

screen shot 2017-02-19 at 16 48 29

### DIFF
--- a/src/modules/build/notifyStatus.js
+++ b/src/modules/build/notifyStatus.js
@@ -1,0 +1,72 @@
+import GitHubAPI from 'github'
+import config from 'config'
+import Build from 'server/models/Build'
+import crashReporter from 'modules/crashReporter/crashReporter'
+
+export class NotifyError extends Error {}
+
+const notifyStatus = async (buildId, { state, description }) => {
+  const build = await Build.query().findById(buildId)
+    .eager('[repository, repository.user, repository.organization, compareScreenshotBucket]')
+
+  if (!build) {
+    throw new NotifyError('Build not found')
+  }
+
+  const [user] = await build.getUsers()
+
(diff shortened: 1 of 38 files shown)


oliviertassinari:
Should we wait to have better authorization handling before merging it or we just sync public repositories? Also does it fix the email `null` overrides? Otherwise, we gonna broke everything.

neoziro:
@oliviertassinari only public repositories are synced. The email null is not yet fixed, I will do it.

neoziro:
@oliviertassinari I don't synchronize the authenticated user (we already do it at login), so I think the email problem is gone.

src/modules/build/notifyStatus.js
```diff
@@ -0,0 +1,72 @@
...
+    token: user.accessToken,
+  })
+
+  return github.repos.createStatus({
+    owner: owner.name,
+    repo: build.repository.name,
+    sha: build.compareScreenshotBucket.commit,
+    state,
+    description,
+    target_url: 'https://www.argos-ci.com/', // TODO compute build url
```
oliviertassinari: We can use `config.get('server.url')` to begin with.

src/server/models/BaseModel.js
```diff
@@ -1,6 +1,11 @@
 import { Model } from 'objection'
 
 export default class BaseModel extends Model {
+  constructor(props) {
+    super(props)
+    Object.assign(this, props)
```
oliviertassinari: Ahaha, le hack

src/modules/build/notifyStatus.js
neoziro: We will change it with the url of the build, for now it's not a problem.

PR #93, opened 2017-02-19
feat(review): start sorting screenshot diffs by DESC score

oliviertassinari:


src/server/models/ScreenshotDiff.test.js
```diff
@@ -0,0 +1,49 @@
...
+  compareScreenshotId: '2',
+  jobStatus: 'pending',
+  validationStatus: 'unknown',
+}
+
+describe('models/ScreenshotDiff', () => {
+  describe('validation score', () => {
+    it('should throw if the score is invalid', () => {
+      assert.throws(() => {
+        ScreenshotDiff.fromJson({
```
neoziro: I miss that! I tried new but without success...

neoziro:
I think you should use http://json-schema.org/latest/json-schema-validation.html for validation.

src/server/models/ScreenshotDiff.test.js
oliviertassinari: Same here, I found it in the tests of objection. It's a public API.

PR #102, opened 2017-02-21
chore: stronger jobs

### LINKED ISSUE #100: Retry at most 3 times a screenshotDiffs task then fail
Stop retrying the jobs after 3 fails. If we fail more than 3 times, we should flag an error.

### Description

Following our postmortem, we shouldn't retry the `screenshotDiffs` job indefinitely.
We made 800k screenshots in a single day and had many errors:
capture d ecran 2017-02-21 a 13 24 28

neoziro:
Closes #100

### DIFF
--- a/src/modules/jobs/createJob.js
+++ b/src/modules/jobs/createJob.js
@@ -1,25 +1,47 @@
 import { getChannel } from 'server/amqp'
 import crashReporter from 'modules/crashReporter/crashReporter'
 
+const serializeMessage = payload => new Buffer(JSON.stringify(payload))
+const parseMessage = (message) => {
+  const payload = JSON.parse(message.toString())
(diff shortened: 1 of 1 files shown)


src/modules/jobs/createJob.js
```diff
@@ -1,25 +1,47 @@
 import { getChannel } from 'server/amqp'
 import crashReporter from 'modules/crashReporter/crashReporter'
 
+const serializeMessage = payload => new Buffer(JSON.stringify(payload))
+const parseMessage = (message) => {
+  const payload = JSON.parse(message.toString())
+  if (!Array.isArray(payload.args) || typeof payload.attempts !== 'number') {
```
oliviertassinari: Can it be an float? Otherwise `Number.isInteger` sounds better.

PR #117, opened 2017-02-22
refactor(jobs): make it generic and notify errors

### LINKED ISSUE #112: Put job in error when it fails
Job should be marked in error in database when it fails 3 times.

neoziro:
Closes #112

### DIFF
--- a/src/modules/build/computeScreenshotDiff.test.js
+++ b/src/modules/build/computeScreenshotDiff.test.js
@@ -62,9 +62,8 @@ describe('computeScreenshotDiff', () => {
     })
 
     await screenshotDiff.reload()
-    expect(screenshotDiff.score > 0).toBeTruthy()
-    expect(screenshotDiff.s3Id).not.toBeUndefined()
-    expect(screenshotDiff.s3Id).not.toBeNull()
+    expect(screenshotDiff.score > 0).toBe(true)
+    expect(typeof screenshotDiff.s3Id === 'string').toBe(true)
     expect(notifyFailure).toBeCalledWith(build.id)
   })
 })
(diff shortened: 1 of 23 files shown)


src/server/routes/configurePassport.js
```diff
@@ -33,6 +31,7 @@ export default (passport) => {
           .insert({
             githubId: Number(profile.id),
             accessToken,
+            login: profile.username,
```
oliviertassinari: Why not doing it in `getDataFromProfile`?

src/server/models/Build.test.js
```diff
@@ -7,6 +7,7 @@ const baseData = {
   repositoryId: '1',
   baseScreenshotBucketId: '1',
   compareScreenshotBucketId: '2',
+  jobStatus: 'pending',
```
oliviertassinari: Why, I don't think we need it.

src/modules/build/computeScreenshotDiff.test.js
```diff
@@ -56,13 +56,15 @@ describe('computeScreenshotDiff', () => {
...
+    await computeScreenshotDiff(screenshotDiff, {
       s3,
       bucket: 'argos-screenshots-sandbox',
     })
-    expect(resultScreenshotDiff.score > 0).toBeTruthy()
-    expect(resultScreenshotDiff.jobStatus).toBe('complete')
-    expect(resultScreenshotDiff.s3Id).not.toBeUndefined()
+
+    await screenshotDiff.reload()
+    expect(screenshotDiff.score > 0).toBeTruthy()
```
oliviertassinari: true

src/server/models/Build.test.js
neoziro: We need it because your use `Build.fromJson`

PR #123, opened 2017-02-26
feat: avoid uploading screenshot if score === 0

### LINKED ISSUE #119: Avoid uploading diffs when the score is 0
### Motivation

Save money

### Detail

It's inside the screenshotDiff job.

neoziro:
Closes #119


src/server/models/ScreenshotDiff.js
```diff
@@ -71,6 +71,16 @@ export default class ScreenshotDiff extends BaseModel {
     },
   };
 
+  $parseDatabaseJson(json) {
+    json = super.$parseDatabaseJson(json)
+
+    if (typeof json.score === 'string') {
```
oliviertassinari: How can this be a string?

neoziro: I don't know probably a bug in the driver, float => string.

PR #126, opened 2017-02-26
feat: notify with build url

### LINKED ISSUE #105: Add a direct link to the build in the GitHub notification
ETQU, Instead of being redirecting to the homepage of argos, I'm redirected to the specific build.

### Motivation

It saves time for users.

### Detail

We gonna need a path builder, we could use rails as inspiration.

neoziro:
Closes #105

Also dramatically speed up test by using delete instead of truncate (x4 speed up).

### DIFF
--- a/config/test.json
+++ b/config/test.json
@@ -1,6 +1,7 @@
 {
   "server": {
     "logFormat": "",
+    "url": "http://www.argos-ci.test"
   },
   "s3": {
     "screenshotsBucket": "argos-screenshots-test"
--- a/src/modules/urls/buildUrl.js
+++ b/src/modules/urls/buildUrl.js
@@ -1,0 +1,17 @@
+import config from 'config'
+
+export async function formatUrlFromBuild(build, { absolute = false } = {}) {
+  if (!build.repository) {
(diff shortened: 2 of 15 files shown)


config/test.json
```diff
@@ -1,6 +1,7 @@
 {
   "server": {
     "logFormat": "",
+    "url": "http://www.argos-ci.test"
```
oliviertassinari: why not using `url`?

src/modules/urls/buildUrl.js
```diff
@@ -0,0 +1,17 @@
+import config from 'config'
+
+export async function formatUrlFromBuild(build, { absolute = false } = {}) {
```
oliviertassinari: Shouldn't we make that module cross environment (browser/node)? I see some places where I could use it on the front.

neoziro: A little complicated, but why not :)

config/test.json
neoziro: It's cleaner I think, the dev url is not very nice and it's not relevant.

oliviertassinari: I mean, the use case it the same. Aren't we introducing a duplicate?

PR #130, opened 2017-02-26
fix(seed): update the seed for the new constraintes

oliviertassinari:


oliviertassinari:
I think that we should be running the seed in the ci test suite.

PR #133, opened 2017-02-27
fix(bdd): max connections reached

oliviertassinari:


knexfile.js
```diff
@@ -29,7 +33,7 @@ const config = {
     client: 'postgresql',
     pool: {
       min: 2,
-      max: 10,
+      max: Math.floor((maxConnectionsAllowed - freeConnectionsForThierdTools) / workers),
```
oliviertassinari: It's going to be `6` here.

PR #141, opened 2017-03-04
chore(database): consolidate db constraints

### LINKED ISSUE #94: Put every constraints into database and into models
It's important to keep data very consistent and this from the beginning, the idea is to migrate database and add model validation to start with consistent data.

neoziro:
Already ran in production

Closes #94

### DIFF
--- a/migrations/20170304184220_add_constraints.js
+++ b/migrations/20170304184220_add_constraints.js
@@ -1,0 +1,39 @@
+/* eslint-disable quotes */
+
+exports.up = knex =>
+  knex.schema
+    .raw(`CREATE TYPE job_status AS ENUM ('pending', 'progress', 'complete', 'error')`)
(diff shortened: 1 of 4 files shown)


migrations/20170304184220_add_constraints.js
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable quotes */
+
+exports.up = knex =>
+  knex.schema
+    .raw(`CREATE TYPE job_status AS ENUM ('pending', 'progress', 'complete', 'error')`)
```
oliviertassinari: I'm wondering if we need PG constraints once we have the objections ones, but why not :).

neoziro: Yes we need, I want to be able to use Postico or psql without the risk of corrupting data.

PR #142, opened 2017-03-04
chore: upgrade node and packages

neoziro:


package.json
```diff
@@ -96,39 +96,39 @@
...
-    "eslint-plugin-babel": "^4.0.1",
+    "eslint-plugin-babel": "^4.1.0",
     "eslint-plugin-import": "^2.1.0",
     "eslint-plugin-jsx-a11y": "^3.0.2",
     "eslint-plugin-mocha": "^4.7.0",
     "eslint-plugin-react": "^6.10.0",
     "factory-girl": "^4.2.1",
     "foreman": "^2.0.0",
-    "jest": "^18.1.0",
+    "jest": "^19.0.2",
```
oliviertassinari: I hope they fixed there memory leak.

neoziro: I found an issue relative to notify, so I disabled it. I think if tests are passing, it's safe.

Issue #59, opened 2017-01-29
Diff UI: validate changes


oliviertassinari:
As a user, I have access to a validation button that is going to accept the changes.

## Motivation

So we can keep track of what was accepted and what wasn't.

## Details

- A graphQL mutation API will be exposed.
- We only one button

PR #152, opened 2017-03-05
feat(review): send the review down to github

oliviertassinari:
For instance https://api.github.com/repos/callemall/material-ui/statuses/8fcaca081dcf18815b474d68b3c4952f4adc83cb?access_token=084fc8f8dfaf25386ed5c1b12cdb9eff4ff735e7.

The next step for #59 is to display the status correctly.

### DIFF
--- a/migrations/20170306205356_new-notifications.js
+++ b/migrations/20170306205356_new-notifications.js
@@ -1,0 +1,26 @@
+/* eslint-disable quotes, max-len */
+
+exports.up = knex =>
+  knex.schema
+    .raw(`ALTER TYPE build_notifications_type ADD VALUE 'diff-accepted'`)
+    .raw(`ALTER TYPE build_notifications_type ADD VALUE 'diff-rejected'`)
+
+exports.down = knex =>
+  knex.schema
+    .raw(`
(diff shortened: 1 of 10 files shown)


oliviertassinari:
I should also add some authorization mechanism.

neoziro:
@oliviertassinari yes some authorization are missing on build.

migrations/20170306205356_new-notifications.js
```diff
@@ -0,0 +1,26 @@
+/* eslint-disable quotes, max-len */
+
+exports.up = knex =>
+  knex.schema
+    .raw(`ALTER TYPE build_notifications_type ADD VALUE 'diff-accepted'`)
+    .raw(`ALTER TYPE build_notifications_type ADD VALUE 'diff-rejected'`)
```
oliviertassinari: That's kind of a pain to keep that up to date. We will see how this evolves.

neoziro: I don't think so. But we can do it using check constraints instead of types. But I think it's very important to have it in database either a check constraints or a type.

PR #172, opened 2017-03-11
chore: redirect from http to https

### LINKED ISSUE #155: Redirect from http to https
As a user, when I use http scheme, I am redirected to https scheme.

### Motivation

Be sure that all users are secured.

### Detail

Redirection should probably done in express because when can't do it directly on heroku.. I think.

neoziro:
Closes #155

### DIFF
--- a/src/server/routes/app.js
+++ b/src/server/routes/app.js
@@ -26,6 +26,18 @@ if (config.get('server.logFormat')) {
 
 app.use(compress())
 app.use(csp) // Content Security Policy
+
+// Redirect from http to https
+if (config.get('server.secure')) {
+  app.use((req, res, next) => {
+    if (req.headers['x-forwarded-proto'] !== 'https') {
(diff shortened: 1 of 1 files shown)


neoziro:
@oliviertassinari actually in production, tested and it works well.

src/server/routes/app.js
```diff
@@ -26,6 +26,18 @@ if (config.get('server.logFormat')) {
 
 app.use(compress())
 app.use(csp) // Content Security Policy
+
+// Redirect from http to https
+if (config.get('server.secure')) {
+  app.use((req, res, next) => {
+    if (req.headers['x-forwarded-proto'] !== 'https') {
```
oliviertassinari: > The X-Forwarded-Proto (XFP) header is a de-facto standard header for identifying the protocol (HTTP or HTTPS) that a client used to connect to your proxy or load balancer.

PR #180, opened 2017-03-13
fix(profile): filter repositories by owner

neoziro:


neoziro:
@oliviertassinari no time, I do it later.

PR #196, opened 2017-03-16
fix(review): small tweeks

### LINKED ISSUE #189: Sort screenshot diffs using an alphanumeric order
That order comes after the score one.

oliviertassinari:
Closes #189.

### DIFF
--- a/src/review/routes/profile/Profile.js
+++ b/src/review/routes/profile/Profile.js
@@ -1,16 +1,19 @@
 import React, { PropTypes } from 'react'
 import { Link as LinkRouter } from 'react-router'
+import { createStyleSheet } from 'jss-theme-reactor'
 import { connect } from 'react-redux'
 import recompact from 'modules/recompact'
 import Text from 'material-ui/Text'
 import Paper from 'material-ui/Paper'
 import Avatar from 'material-ui/Avatar'
 import Layout from 'material-ui/Layout'
+import { white } from 'material-ui/styles/colors'
 import {
   List,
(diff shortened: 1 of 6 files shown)


src/review/routes/profile/Profile.js
```diff
@@ -42,17 +54,15 @@ function Profile(props) {
               <Layout item>
                 <Avatar
                   src={`https://github.com/${profileName}.png?size=300`}
-                  style={{ width: 120, height: 120 }}
+                  className={classes.avatar}
```
neoziro: J'avais eu la flemme :x

src/server/graphql/BuildType.js
```diff
@@ -60,7 +60,11 @@ const BuildType = new GraphQLObjectType({
     screenshotDiffs: {
       description: 'Get the diffs for a given build.',
       type: new GraphQLList(ScreenshotDiffType),
-      resolve: build => build.$relatedQuery('screenshotDiffs').orderBy('score', 'desc'),
+      resolve: source => source
+        .$relatedQuery('screenshotDiffs')
+        .innerJoin('screenshots', 'screenshots.id', 'screenshot_diffs.baseScreenshotId')
```
oliviertassinari: We have quite some N+1 query with that one. That's quick and dirty.

neoziro: Yes, but it can be easily fixed with GraphQL. If the performance is correct, sounds good for me.

PR #210, opened 2017-03-19
fix: handle authorization correctly

### LINKED ISSUE #197: Check user rights on Argos
As a user, I have a warning if my token doesn't match with my rights "privateSync" status in "users" table. 

### Motivation

Prevent users to be stuck.

neoziro:
screen shot 2017-03-19 at 14 27 51

Fix #197

### DIFF
--- a/src/review/modules/components/AuthorizationNotice.js
+++ b/src/review/modules/components/AuthorizationNotice.js
@@ -1,0 +1,56 @@
+import React, { PropTypes } from 'react'
+import { connect } from 'react-redux'
+import { createStyleSheet } from 'jss-theme-reactor'
+import withStyles from 'material-ui/styles/withStyles'
+import recompact from 'modules/recompact'
+import Paper from 'material-ui/Paper'
+import Layout from 'material-ui/Layout'
+import Button from 'material-ui/Button'
+import Link from 'modules/components/Link'
+import authorizationStatuses, { CONSISTENT } from 'modules/authorizations/authorizationStatuses'
+
+const styleSheet = createStyleSheet('AuthorizationNotice', () => ({
+  paper: {
+    padding: 16,
+  },
+}))
+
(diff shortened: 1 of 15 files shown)


src/review/modules/components/AuthorizationNotice.js
```diff
@@ -0,0 +1,56 @@
...
+      <Layout container align="center" justify="center">
+        <Layout item xs={12} sm>
+          Your GitHub authentification is outdated, please authenticate to fix it.
+        </Layout>
+        <Layout item sm={false}>
+          <Button
+            accent
+            compact
+            component={Link}
+            href={user.privateSync ? '/auth/github-private' : '/auth/github-public'}
```
oliviertassinari: Should we have a pathname mapping somewhere like we have with rails?

src/review/modules/components/AuthorizationNotice.js
```diff
@@ -0,0 +1,56 @@
...
+            raised
+          >
+            {'Authenticate to GitHub'}
```
oliviertassinari: Useless complexity 😋

src/review/modules/components/AuthorizationNotice.js
```diff
@@ -0,0 +1,56 @@
+import React, { PropTypes } from 'react'
+import { connect } from 'react-redux'
+import { createStyleSheet } from 'jss-theme-reactor'
+import withStyles from 'material-ui/styles/withStyles'
```
oliviertassinari: I have made an update to merge those two dependencies but it's not released yet.
```jsx
import { createStyleSheet, withStyles } from 'material-ui/styles'
```

src/review/modules/components/AuthorizationNotice.js
neoziro: Why not :)

PR #212, opened 2017-03-19
style: remove useless complexity

neoziro:


neoziro:
@oliviertassinari and probably never, as GitHub

oliviertassinari:
We are also targeting designers in the review process, we will see.

PR #215, opened 2017-03-19
Update README.md

oliviertassinari:
Let's see if argos can handle that.


oliviertassinari:
We don't run the CI with external PRs. Well, at lease our env variables are safe.

PR #245, opened 2017-04-01
chore: improve error handler

### LINKED ISSUE #242: Put an HTML error handler
As a user, I can see a beautiful error page.

### Motivation

JSON is not very user friendly you know...

### Detail

Very easy to do it using [express-err](https://github.com/neoziro/express-err/).

neoziro:
Closes #242 

screen shot 2017-04-01 at 18 19 51

Message and stack trace are not displayed in production.

### DIFF
--- a/src/review/routes/Error.js
+++ b/src/review/routes/Error.js
@@ -1,0 +1,44 @@
+import React, { PropTypes } from 'react'
+import { compose } from 'recompact'
+import { connect } from 'react-redux'
+import { withStyles, createStyleSheet } from 'material-ui/styles'
+import LayoutBody from 'modules/components/LayoutBody'
+import ViewContainer from 'modules/components/ViewContainer'
+import ScrollView from 'modules/components/ScrollView'
+import ReviewAppBar from 'review/modules/AppBar/AppBar'
+
+const styleSheet = createStyleSheet('Error', () => ({
+  statusCode: {
+    fontSize: 30,
+    lineHeight: 2,
+  },
+}))
+
+function ErrorView({ classes, error }) {
+  return (
+    <ViewContainer>
+      <ReviewAppBar />
+      <ScrollView>
+        <LayoutBody margin bottom={false}>
+          { <div className={classes.statusCode}>Error {error.statusCode}</div> }
+          { error.message && <div>{error.message}</div> }
+          { error.stack && <code><pre>{error.stack}</pre></code> }
+        </LayoutBody>
+      </ScrollView>
+    </ViewContainer>
+  )
+}
+
+ErrorView.propTypes = {
+  classes: PropTypes.object,
+  error: PropTypes.shape({
+    message: PropTypes.string,
+    stack: PropTypes.string,
+    statusCode: PropTypes.number.isRequired,
+  }),
+}
+
+export default compose(
+  withStyles(styleSheet),
+  connect(state => ({ error: state.data.error })),
+)(ErrorView)
--- a/src/review/routes/homepage/Homepage.js
+++ b/src/review/routes/homepage/Homepage.js
@@ -1,13 +1,23 @@
 import React, { PropTypes } from 'react'
 import { connect } from 'react-redux'
 import ProductHome from 'review/routes/product/Home'
+import Error from 'review/routes/Error'
 import Dashboard from 'review/routes/dashboard/Dashboard'
 
-function Homepage(props) {
-  return props.user ? <Dashboard /> : <ProductHome />
+function Homepage({ user, error }) {
+  if (error) {
+    return <Error />
+  }
+
+  if (user) {
+    return <Dashboard />
+  }
+
+  return <ProductHome />
 }
 
 Homepage.propTypes = {
+  error: PropTypes.object,
   user: PropTypes.object,
 }
 
--- a/src/server/middlewares/rendering.js
+++ b/src/server/middlewares/rendering.js
@@ -52,7 +52,7 @@ function injectJSON(data) {
   return JSON.stringify(data, null, process.env.NODE_ENV === 'production' ? 0 : 2)
 }
 
-export default (req, res) => {
+export default additionalClientData => (req, res) => {
   const output = ejs.render(indexString, {
     cache: true,
     filename: 'review/index.ejs',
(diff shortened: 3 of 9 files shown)


src/review/routes/Error.js
```diff
@@ -0,0 +1,41 @@
+import React, { PropTypes } from 'react'
+import { compose } from 'recompact'
+import { connect } from 'react-redux'
+import { withStyles, createStyleSheet } from 'material-ui/styles'
+import Paper from 'material-ui/Paper'
+import ViewContainer from 'modules/components/ViewContainer'
+import ScrollView from 'modules/components/ScrollView'
+import ReviewAppBar from 'review/modules/AppBar/AppBar'
+
+const styleSheet = createStyleSheet('ProductHome', () => ({
```
oliviertassinari: `'Error'`

src/review/routes/Error.js
```diff
@@ -0,0 +1,41 @@
...
+  landing: {
+    padding: 15,
+  },
+}))
+
+const Error = ({ classes, error }) => (
+  <ViewContainer>
+    <ReviewAppBar />
+    <ScrollView>
+      <Paper rounded={false} elevation={3} className={classes.landing}>
```
oliviertassinari: Why not using the `LayoutBody`?

src/review/routes/Error.js
```diff
@@ -0,0 +1,41 @@
...
+        { <div style={{ fontSize: '30px', lineHeight: 2 }}>Error {error.statusCode}</div> }
```
oliviertassinari: `fontSize: 30,`

oliviertassinari: CSS?

src/review/routes/Error.js
```diff
@@ -0,0 +1,41 @@
...
+  error: PropTypes.shape({
+    message: PropTypes.string,
+    stack: PropTypes.string,
+    statusCode: PropTypes.number.isRequired,
+  }),
+}
+
+export default compose(
+  withStyles(styleSheet),
+  connect(({ data: { error } }) => ({ error })),
```
oliviertassinari:
```js
connect((state) => ({ error: state.data.error })),
```
?
Looks much better for grepping

src/review/routes/Error.js
```diff
@@ -0,0 +1,41 @@
...
+import ScrollView from 'modules/components/ScrollView'
+import ReviewAppBar from 'review/modules/AppBar/AppBar'
+
+const styleSheet = createStyleSheet('ProductHome', () => ({
+  landing: {
+    padding: 15,
+  },
+}))
+
+const Error = ({ classes, error }) => (
```
oliviertassinari: function to be consistent

src/review/routes/homepage/Homepage.js
```diff
@@ -1,13 +1,23 @@
 import React, { PropTypes } from 'react'
 import { connect } from 'react-redux'
 import ProductHome from 'review/routes/product/Home'
+import Error from 'review/routes/Error'
 import Dashboard from 'review/routes/dashboard/Dashboard'
 
-function Homepage(props) {
-  return props.user ? <Dashboard /> : <ProductHome />
+function Homepage({ user, error }) {
```
oliviertassinari: I would rather use `props` to be consistent.

src/server/middlewares/rendering.js
```diff
@@ -52,7 +52,7 @@ function injectJSON(data) {
   return JSON.stringify(data, null, process.env.NODE_ENV === 'production' ? 0 : 2)
 }
 
-export default (req, res) => {
+export default additionalClientData => (req, res) => {
```
oliviertassinari: What about moving `additionalClientData` into the `response`?

src/review/routes/homepage/Homepage.js
oliviertassinari: Actually, this pattern is breaking https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prop-types.md.

src/review/routes/Error.js
neoziro: Ok

src/review/routes/Error.js
neoziro: Less code is better?

src/review/routes/Error.js
neoziro: Yes I should use it.

src/review/routes/Error.js
neoziro: Yes, but not a problem for now.

src/review/routes/Error.js
neoziro: What grepping?

src/review/routes/homepage/Homepage.js
neoziro: No it doesn't break it I had the warning. I think destructuring is more explicit and dryer.

src/server/middlewares/rendering.js
neoziro: I don't understand, if it's mutating response, I think it's not a good idea.

src/review/routes/Error.js
neoziro: You should grep "connect(".

src/review/routes/homepage/Homepage.js
oliviertassinari: Right, it's working.

src/review/routes/Error.js
oliviertassinari: I don't know, it feels more idiomatic without the destructuration.

oliviertassinari: by grepping, I'm referring to looking for where the `state.data` is used.

src/review/routes/Error.js
oliviertassinari: From my point of view, flexibility and consistency are serious to less code.

src/server/middlewares/rendering.js
oliviertassinari: Don't we already do that with `req.user`? We could mutate the request otherwise 🙈

neoziro: Passport do that yes, but I don't agree with it.

PR #250, opened 2017-04-02
feat(diff): add a baselineBranch option

### LINKED ISSUE #247: Configuration of the default branch to run diffs against
As a user, I can configure per repository the default branch the tests will be done against.

### Motivation

Material-UI use `next` as the default branch.

oliviertassinari:
Closes #247

### DIFF
--- a/src/modules/synchronizer/GitHubSynchronizer.js
+++ b/src/modules/synchronizer/GitHubSynchronizer.js
@@ -44,6 +44,7 @@ class GitHubSynchronizer {
       } else {
         repository = await Repository.query().insert({
           ...data,
+          baselineBranch: 'master',
           enabled: false,
         })
       }
--- a/src/server/models/Repository.js
+++ b/src/server/models/Repository.js
@@ -10,6 +10,7 @@ export default class Repository extends BaseModel {
       'name',
       'enabled',
       'private',
+      'baselineBranch',
     ],
     properties: {
       githubId: { type: 'number' },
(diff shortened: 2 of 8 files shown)


src/server/models/Repository.js
```diff
@@ -10,6 +10,7 @@ export default class Repository extends BaseModel {
       'name',
       'enabled',
       'private',
+      'baselineBranch',
```
neoziro: not required if default

src/modules/synchronizer/GitHubSynchronizer.js
```diff
@@ -44,6 +44,7 @@ class GitHubSynchronizer {
       } else {
         repository = await Repository.query().insert({
           ...data,
+          baselineBranch: 'master',
```
neoziro: Only one default please

src/server/test/factory.js
```diff
@@ -62,6 +62,7 @@ factory.define('Repository', Repository, {
   githubId: factory.sequence('repository.githubId', n => n),
   name: factory.sequence('repository.name', n => `repo-${n}`),
   enabled: true,
+  baselineBranch: 'master',
```
neoziro: Only one default

neoziro:
I understand your point of view, but for me it's dangerous because if you forgot to change the database one the day you change the default, your application is no longer coherent. In fact it increases the change of having an incoherent state in the database.

neoziro:
@oliviertassinari for this example is not very true, but in general I think it is.

oliviertassinari:
I'm gonna remove the default databse

neoziro:
Good for me.

PR #253, opened 2017-04-04
chore: simplify eslint

neoziro:
Also be drier by removing code and prepare eslint-config-argos.


neoziro:
Long term vision is to dry and simplify all front code to be able to developer make x3 on dev time.

src/review/routes/Error.js
```diff
@@ -1,4 +1,4 @@
-import React, { PropTypes } from 'react'
+import React from 'react'
```
neoziro: In this file -13 lines of code for the same thing. We can code it twice faster.

src/review/routes/Error.js
```diff
@@ -14,31 +14,18 @@ const styleSheet = createStyleSheet('Error', () => ({
...
-    stack: PropTypes.string,
-    statusCode: PropTypes.number.isRequired,
-  }),
-}
-
 export default compose(
   withStyles(styleSheet),
   connect(state => ({ error: state.data.error })),
-)(ErrorView)
+)(({ classes, error }) => (
```
oliviertassinari: I don't get the value if this change.
- we lost the idiomatic display name
- we break the consistancy, some time the HOC are under the component (class) somethimes above (function).

src/review/routes/build/BuildSummary.js
```diff
@@ -36,12 +34,12 @@ export function BuildSummary(props) {
...
+BuildSummaryView.propTypes = {
   classes: PropTypes.object.isRequired,
   fetch: PropTypes.object.isRequired,
 }
 
 export default recompact.compose(
   withStyles(styleSheet),
   connect(state => state.ui.build),
-)(BuildSummary)
+)(BuildSummaryView)
```
oliviertassinari: Can we rename the file and stylesheet name? I would rather have a strong mapping between component names and files.

.eslintrc.js
```diff
@@ -1,78 +1,38 @@
...
-    'jsx-a11y',
-    'mocha',
-  ],
+  env: {
+    jest: true,
+    browser: true,
+  },
   rules: {
-    'max-len': ['error', 100, 2, {
+    'max-len': ['error', 120, 2, {
```
oliviertassinari: https://www.quora.com/Why-does-Google-Java-coding-style-guide-chose-maximum-of-100-characters-column-limit-per-line-why-not-120-just-curious-to-know

src/review/routes/Error.js
neoziro: - we don't care about displayName, did you already use it since Argos dev has started?
- consistency is respected for me:

´´´js
export default 'bar'
// I have to decorate it
export default transform('bar')
´´´

The consistency you mean is to write:

´´´js
const foo = 'foo'
export default foo
// I have to decorate it
const foo = 'foo'
export default transform(foo)
´´´

The consistence is not respected for me, you just declare a useless variable.

oliviertassinari: > did you already use it since Argos dev has started?

I do when debugging a React warning, or using the dev tool to understand what's going on.

src/review/routes/build/BuildSummary.js
neoziro: There is a rule about import/no-default-export-name" and I think it's important. Also I think we should not export something just for testing. We actually don't test the thing we export, the test has no value for me.

.eslintrc.js
neoziro: Just a mistake, I aggree for 100.

src/review/routes/build/BuildSummary.js
oliviertassinari: I see 👍

src/review/routes/Error.js
neoziro: OK, it's a debug related thing, I think it should be done automatically by a webpack loader, I will give it a look.

neoziro:
@oliviertassinari I think we could add some rules if you want or stay Airbnb like, we could talk about it today.

PR #254, opened 2017-04-04
feat: 100 commits to compare

neoziro:


neoziro:
@oliviertassinari it costs nothing to go to get 100 commits instead of 30.

oliviertassinari:
Good call, I have noticed a build today that couldn't find the fork commit. Also, I think that we could improve the algorithm when it's not found.

PR #276, opened 2017-05-01
fix(back): security issue

oliviertassinari:
We are leaking the token of Material-UI and Argos-CI, not good 😬 .


src/server/graphql/RepositoryType.js
```diff
@@ -105,6 +108,13 @@ const RepositoryType = new GraphQLObjectType({
     },
     token: {
       type: GraphQLString,
+      resolve: async (repository, args, context) => {
+        if (!await repository.authorization(context.user)) {
```
oliviertassinari: Here is the leak.

oliviertassinari:
Yes.

src/server/graphql/RepositoryType.js
neoziro: OK good catch ;)

Issue #137, opened 2017-02-28
Improve credibility of the homepage


oliviertassinari:
As a user, I can see the following:
- New wordings
  - `Automated visual diff and regression testing`?
- a contact email address
- a twitter account
- a quick overview of the interface

Some inspirations : https://colorlib.com/wp/landing-page-wordpress-themes/

PR #283, opened 2017-05-09
feat(product): Improve credibility of the homepage v2

oliviertassinari:
Keep working on #137.


src/review/modules/components/ProductFooter.js
```diff
@@ -0,0 +1,101 @@
...
+  },
+  icon: {
+    width: 16,
+    height: 16,
+    marginRight: theme.spacing.unit,
+  },
+}))
+
+function ProductFooter(props) {
+  const {
```
neoziro: You should stop ✋

src/review/modules/components/ProductHeader.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { withStyles, createStyleSheet } from 'material-ui/styles'
+import Typography from 'material-ui/Typography'
+import Paper from 'material-ui/Paper'
+import LayoutBody from 'modules/components/LayoutBody'
+
+const styleSheet = createStyleSheet('ProductHeader', theme => ({
```
neoziro: Where is the name 'ProductHeader' used?

oliviertassinari: Just for debugging the stylesheets. I can check if it's required.

src/review/modules/components/ProductFooter.js
oliviertassinari: I like that pattern.
- Already used everywhere, if we change it, we need to do it everywhere.
- Easier to refactor into a class
- Now, I'm used to it with Material-UI

src/review/modules/components/ProductHeader.js
neoziro: I think it's redundant with component displayName, we can just create a hoc that deduce it from the next component exactly like recompose / recompact does.

src/review/modules/components/ProductFooter.js
neoziro: 1. Yes for sure.
2. Refactoring in a class is a bad smell for me, also appears in 5% of cases, VS 95% of cases. I prefer to keep it dry in 95% of cases.
3. OK

The biggest problem for me is the lose of meaning in the signature, you can't answer this question by reading the signature of the function:

- What is the type of `props`
- Is it required?
- OK I assume it's an object, what does it contains?

Destructuring could be optimized in JS engines (OK this function need an object that is required).

PR #312, opened 2017-06-03
docs(with-rails): add a new integration example

oliviertassinari:


examples/with-rails/test/ui_tweak_helper.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+
+module UiTweakHelper
```
oliviertassinari: @neoziro Ton petit bijou

oliviertassinari:
Stylé rails !

```rb
          def display_image
            message = "[Screenshot]: #{image_path}\n"

            case output_type
            when "artifact"
              message << "\e]1338;url=artifact://#{image_path}\a\n"
            when "inline"
              name = inline_base64(File.basename(image_path))
              image = inline_base64(File.read(image_path))
              message << "\e]1337;File=name=#{name};height=400px;inline=1:#{image}\a\n"
            end

            message
          end
```
[*Source*](https://github.com/eileencodes/rails/blob/master/actionpack/lib/action_dispatch/system_testing/test_helpers/screenshot_helper.rb#L65-L78)

examples/with-rails/test/ui_tweak_helper.rb
neoziro: @oliviertassinari ;)

PR #315, opened 2017-06-03
chore(js): upgrade the dependencies

oliviertassinari:


neoziro:
We should use `eslint-config-prettier`.

oliviertassinari:
Isn't overkill? Don't we just need to disable the rule that we see have conflict?

neoziro:
This is exactly the goal of the preset 😂

oliviertassinari:
🤣

PR #328, opened 2017-06-30
Fix production

oliviertassinari:
node v8 isn't ready yet.


neoziro:
@oliviertassinari what was the problem?

oliviertassinari:
node-foreman do not support it. Might have to use pm2 in dev

neoziro:
Or waiting for it to support node v8+.

PR #341, opened 2017-08-05
core(js): upgrade node version

oliviertassinari:


neoziro:
@oliviertassinari why pm2 in dev and foreman in production?

oliviertassinari:
foreman-node doesn't work with node@8 and we are only using forman in production because we are using heroku.

PR #365, opened 2019-04-23
Fix race condition when incrementing batchCount

agaryen:
Some Argos builds were forever pending because of a race condition in the way we incremented the batch counter.
There was a quick time gap between the fetching of the current batchCount and the update of the column.


neoziro:
Thanks!

PR #417, opened 2020-03-29
Allow to use redis instead of AMQP

bpaquet:
At the moment, Argos is not using any features related to AMQP, so we can replace it by redis rpush / lpop.
This PR allows to do this just by changing the CLOUDAMP_URL. If we want to definitely replace AMQP per redis, there are multiple others simplifications we can do.


apps/job-core/src/amqp.js
```diff
@@ -14,8 +15,13 @@ async function connect() {
 let channel
 export async function getAmqpChannel() {
   if (!channel) {
-    const connection = await connect()
-    channel = await connection.createChannel()
+    if (config.get('amqp.url').startsWith('redis://')) {
+      channel = buildAMQPToRedis(config.get('amqp.url'))
```
Izzette: Does it really make sense to re-use this configuration parameter?  What about `mq.type` (default `ampq`) and `redis.url`?  Are there technical challenges to doing this?

apps/job-core/src/amqp_test.js
```diff
@@ -0,0 +1,42 @@
...
+    await sleep(1000)
+    channel.nack(msg, false, false)
+    counter += 1
+    if (counter == 3) {
+      await channel.close()
+    }
+  })
+}
+
+run()
```
Izzette: Should this be a different PR?

apps/job-core/src/amqp_to_redis.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  queueName(queue) {
+    return 'queue_' + queue
+  }
+
+  ack(message) {
+    const callback = this.ack_callbacks[message.id]
+    if (callback) {
+      this.ack_callbacks[message.id] = null
```
Izzette: Does this every get `delete`-ed?  Or will this array just fill up with nulls until it crashes from OOM?

Izzette: Can we simply:
```suggestion
      delete this.ack_callbacks[message.id]
```

apps/job-core/src/amqp_to_redis.js
```diff
@@ -0,0 +1,76 @@
...
+      callback()
```
Izzette: Should this be done synchronously?  This will block other messages from being received or processed as the callback is running.  Is this desirable behaviour or am I missing something?

apps/job-core/src/amqp_to_redis.js
```diff
@@ -0,0 +1,76 @@
...
+    }
+  }
+
+  nack(message, _allUpTo = false, _requeue = false) {
```
Izzette: Are `_allUpTo` or `_requeue` used externally?

apps/job-core/src/amqp_to_redis.js
```diff
@@ -0,0 +1,76 @@
...
+    this.ack(message)
+  }
+
+  sendToQueue(queue, message, _opts) {
```
Izzette: Here too, is `_opts` used externally?

apps/job-core/src/amqp.js
bpaquet: Waiting a decision from @gregberge. If we fully move t Redis, tes we will improve the config

apps/job-core/src/amqp_test.js
bpaquet: Good question. Waiting a review

apps/job-core/src/amqp_to_redis.js
bpaquet: No risk of OOM, because the app is written for AMQP

apps/job-core/src/amqp_to_redis.js
bpaquet: Nope, we are duplicating the behaviour of the AMQP API.

apps/job-core/src/amqp_to_redis.js
bpaquet: AMQP API :)

apps/job-core/src/amqp_to_redis.js
bpaquet: AMQP API :)

apps/job-core/src/amqp_to_redis.js
Izzette: Generally parameters starting with `_` are private, so I'm not quite sure they're used as part of the AMQP API.  Moreover, in javascript, you don't need to declare this parameters in order to permit other code to pass them.  Since you're not used, you can just delete them with no side-effects.

```
 0 ✓ isabelle.cowan-bergman@mbp-de-isabelle.home oauth2_proxy $ node
Welcome to Node.js v12.13.0.
Type ".help" for more information.
> function foo(blah) {
... console.log(blah)
... }
undefined
> foo("blah")
blah
undefined
> foo("blah", "extra")
blah
undefined
> foo("blah", extra => "extra")
blah
undefined
```

apps/job-core/src/amqp_to_redis.js
Izzette: I'm not sure I understand.  Maybe the maintainers will be able to shed some light.

apps/job-core/src/amqp_to_redis.js
Izzette: That's all I need to hear 😊

apps/job-core/src/amqp_to_redis.js
```diff
@@ -0,0 +1,76 @@
...
+      else {
+        await this.exec_message(res, message_callback)
+      }
+    }
+  }
+}
+
+export const buildAMQPToRedis = (redis_url) => {
+  return new AMQPToRedis(redis_url)
+}
```
Izzette: Is the no-trailing-newline thing in all the files for this project?

PR #420, opened 2020-06-30
Hide passing screenshots by default

isc:
Hi Greg,

On our builds for Doctolib, we now have something like 3k screenshots and the page takes a while to load. Some of the time is spent waiting for the server payload to arrive but it seems there is also quite some time spent client side building the UI once the payload has arrived. The loading animation actually freezes on my computer. This PR hopes to alleviate that by hiding passing screenshots by default (while still providing a link to toggle them). In my experience using Argos, passing screenshots are rarely useful, that would be one extra click for the rare cases when you need them.
What do you think?


gregberge:
Hello @isc, I know the UX is not very good for a large number of screenshots. Your solution is a quick-win and solve your problem, so OK for me!

PR #430, opened 2020-10-30
x3 more workers

oliviertassinari:
Process screenshots in parallel


gregberge:
Ah yes we don't have the button to scale up... So let's do this for now.

PR #447, opened 2022-08-03
feat: add plans and purchases tables

jsfez:
@gregberge I need your review on `purchase` table schema


apps/database/migrations/20220803095315_add_plans.js
```diff
@@ -0,0 +1,35 @@
+exports.up = async (knex) => {
+  await knex.schema.createTable('plans', (table) => {
+    table.bigincrements('id').primary()
+    table.dateTime('createdAt').notNullable()
+    table.dateTime('updatedAt').notNullable()
```
gregberge, line 5: I think there is a timestamps option to create updatedAt/createdAt at the same time but fine.

apps/database/migrations/20220803095315_add_plans.js
```diff
@@ -0,0 +1,37 @@
...
+    table.string('name').notNullable()
+    table.float('monthlyPrice').notNullable()
+    table.float('annualPrice').notNullable()
```
gregberge, line 8: What is it?

apps/database/migrations/20220803095315_add_plans.js
```diff
@@ -0,0 +1,37 @@
+exports.up = async (knex) => {
+  await knex.schema.createTable('plans', (table) => {
+    table.bigincrements('id').primary()
+    table.dateTime('createdAt').notNullable()
+    table.dateTime('updatedAt').notNullable()
+    table.string('name').notNullable()
```
gregberge, line 6: Pricing are not useful, please remove it.

apps/database/migrations/20220803095315_add_plans.js
```diff
@@ -0,0 +1,37 @@
...
+    table.bigInteger('userId').index()
+    table.foreign('userId').references('users.id')
+
+    table.bigInteger('organizationId').index()
+    table.foreign('organizationId').references('organizations.id')
+
+    table.string('event').notNullable()
+
+    table.bigInteger('planId').notNullable().references('plans.id')
+    table.foreign('planId').references('plans.id')
```
gregberge, line 28: ??

gregberge (changes requested):
We talk about subscriptions not purchases. We need a start date to know when to start a plan to count the screenshot for the current month? Else we should price a smaller amount for the first month. I don't know what is possible with GitHub.

apps/database/migrations/20220803095315_add_plans.js
jsfez: `table.timestamps(true, true)` but never used in the project until now

apps/database/migrations/20220803095315_add_plans.js
```diff
@@ -0,0 +1,33 @@
...
+    table.integer('screenshotsQuota').notNullable()
+    table.string('githubId')
+  })
+
+  await knex.schema.createTable('purchases', (table) => {
+    table.bigincrements('id').primary()
+    table.timestamps(true, true)
+    table.bigInteger('planId').index()
+    table.foreign('planId').references('plans.id')
+    table.bigInteger('accountId').index()
```
gregberge, line 24: not null

apps/database/migrations/20220803095315_add_plans.js
```diff
@@ -0,0 +1,33 @@
...
+    table.timestamps(true, true)
+    table.string('name').notNullable()
+    table.integer('screenshotsQuota').notNullable()
+    table.string('githubId')
+  })
+
+  await knex.schema.createTable('purchases', (table) => {
+    table.bigincrements('id').primary()
+    table.timestamps(true, true)
+    table.bigInteger('planId').index()
```
gregberge, line 22: not null

apps/database/migrations/20220803095315_add_plans.js
```diff
@@ -0,0 +1,33 @@
...
+    table.bigInteger('organizationId').index()
+    table.foreign('organizationId').references('organizations.id')
+  })
+
+  await knex.schema.createTable('plans', (table) => {
+    table.bigincrements('id').primary()
+    table.timestamps(true, true)
+    table.string('name').notNullable()
+    table.integer('screenshotsQuota').notNullable()
+    table.string('githubId')
```
gregberge, line 16: not null

apps/database/migrations/20220803095315_add_plans.js
```diff
@@ -0,0 +1,33 @@
+exports.up = async (knex) => {
+  await knex.schema.createTable('accounts', (table) => {
+    table.bigincrements('id').primary()
+    table.timestamps(true, true)
+    table.bigInteger('userId').index()
+    table.foreign('userId').references('users.id')
+    table.bigInteger('organizationId').index()
+    table.foreign('organizationId').references('organizations.id')
```
gregberge, line 8: constraint, only userId or organizationId

apps/database/src/models/Account.js
```diff
@@ -0,0 +1,39 @@
...
+  static get tableName() {
+    return 'accounts'
+  }
+
+  static get jsonSchema() {
+    return mergeSchemas(timestampsSchema, {
+      properties: {
+        userId: { type: ['string', null] },
+        organizationId: { type: ['string', null] },
+      },
```
gregberge, line 15: one of with userId or organizationId null / required ?

`oneOf` is possible.

gregberge (changes requested):
We will add GraphQL schema if needed.

gregberge (changes requested):
This PR is complicated because it does not stick to a user story, without a real use case implementing a database is difficult. You should focus on the first one, got the event and link the user.

apps/database/src/models/Account.js
jsfez: I can't find a doc or an example about the `oneOf` options.

jsfez:
> This PR is complicated because it does not stick to a user story, without a real use case implementing a database is difficult. You should focus on the first one, got the event and link the account with the plan.

I update the PR scope. The main purpose is to register a subscription in our database when the matching event is received.

apps/synchronize/src/helpers.js
```diff
@@ -20,3 +27,77 @@ export async function synchronizeFromUserId(userId) {
...
+  const newPurchase = await Purchase.query().insertAndFetch({
+    accountId,
+    planId,
+  })
+  return newPurchase
+}
+
+export async function getMatchingPlan(payload) {
+  const { id: planGithubId } = payload.marketplace_purchase.plan
+  const plan = await Plan.query().where({ githubId: planGithubId }).first()
```
gregberge, line 98:
```suggestion
  const plan = await Plan.query().findOne({ githubId: planGithubId })
```

apps/synchronize/src/helpers.js
```diff
@@ -20,3 +27,77 @@ export async function synchronizeFromUserId(userId) {
...
+  }
+
+  const organization = await getOrCreateOrganization(payload)
+  return Account.query().insertAndFetch({ organizationId: organization.id })
+}
+
+export async function getOrCreatePurchase({ accountId, planId }) {
+  const purchase = await Purchase.query().where({ accountId, planId }).first()
+  if (purchase) return purchase
+  const newPurchase = await Purchase.query().insertAndFetch({
```
gregberge, line 89:
```suggestion
  return Purchase.query().insertAndFetch({
```

apps/synchronize/src/helpers.js
```diff
@@ -20,3 +27,77 @@ export async function synchronizeFromUserId(userId) {
...
+    const user = await getOrCreateUser(payload)
+    return Account.query().insertAndFetch({ userId: user.id })
+  }
+
+  const organization = await getOrCreateOrganization(payload)
+  return Account.query().insertAndFetch({ organizationId: organization.id })
+}
+
+export async function getOrCreatePurchase({ accountId, planId }) {
+  const purchase = await Purchase.query().where({ accountId, planId }).first()
```
gregberge, line 87:
```suggestion
  const purchase = await Purchase.query().findOne({ accountId, planId })
```

apps/database/migrations/20220803095315_add_plans.js
```diff
@@ -0,0 +1,38 @@
...
+  await knex.raw(
+    'ALTER TABLE accounts ADD CONSTRAINT accounts_only_one_non_null CHECK ("userId" IS NOT NULL OR "organizationId" IS NOT NULL)',
+  )
+
+  await knex.schema.createTable('plans', (table) => {
+    table.bigincrements('id').primary()
+    table.dateTime('createdAt').notNullable()
+    table.dateTime('updatedAt').notNullable()
+    table.string('name').notNullable()
+    table.integer('screenshotsQuota').notNullable()
```
gregberge, line 20: screenshotLimitPerMonth ? Parce que là c'est pas très explicit.

gregberge (changes requested):
Je ne comprend pas trop ce que viennent faire les "purchases" dans l'app "synchronize".

PR #458, opened 2022-08-11
feat: logger throws prod errors

jsfez:
I need to monitor production errors on Sentry. The previous logger hides the errors.
@gregberge can you confirm if that's a good fix?

### DIFF
--- a/apps/logger/src/index.js
+++ b/apps/logger/src/index.js
@@ -1,4 +1,6 @@
 /* eslint-disable no-console */
+import * as Sentry from "@sentry/node";
+
 const logger = {
   info: (...args) => {
     if (process.env.NODE_ENV === "test") {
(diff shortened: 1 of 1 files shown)


gregberge (changes requested):
No it is not, the whole process would crash. It is probably better to do Sentry.catchMessage here? Or Sentry.catchException?

jsfez:
> No it is not, the whole process would crash.

That would be annoying. I use the `Sentry.catchException` as suggested.

apps/logger/src/index.js
```diff
@@ -12,6 +14,7 @@ const logger = {
       throw new Error(args);
     }
 
+    Sentry.captureException(args);
```
gregberge, line 17: `args` is an array, not an error. You should do:

```js
if (args[0] instanceof Error) {
  Sentry.captureException(args[0])
} else if (typeof args[0] === 'string') {
  Sentry.captureMessage(args[0])
}

gregberge (changes requested):

PR #482, opened 2022-08-31
refactor: update owner definition

jsfez:


apps/app/src/containers/Apollo.js
```diff
@@ -17,6 +17,11 @@ function ApolloProvider({ children, authToken }) {
         headers: {
           authorization,
         },
+        defaultOptions: {
+          watchQuery: {
+            fetchPolicy: "cache-and-network",
+          },
+        },
```
gregberge, line 24: Remove it please.

apps/config/src/index.js
```diff
@@ -96,7 +96,7 @@ const config = convict({
     appId: {
       doc: "App ID",
       format: String,
-      default: "",
+      default: "57578",
```
gregberge, line 99: Do not edit this file any more please.

gregberge (changes requested):

gregberge (changes requested):
It is not a "feat" but a "refactor". We will generate changelog, it is important.

PR #487, opened 2022-09-02
fix: set apollo owner possible types

jsfez:
fixes ARGOS-BROWSER-3X
fixes 3554264367


apps/app/src/modules/utils.js
```diff
@@ -1,4 +1,6 @@
 /* eslint-disable react/no-unescaped-entities */
 
-export const getPossessiveForm = (str) =>
+export const getPossessiveForm = (str) => {
+  if (!str) return "";
```
gregberge, line 4: Remove it.

gregberge (changes requested):

PR #489, opened 2022-09-02
feat: catch unknown installation id error

jsfez:
fixes ARG-98
fixes ARGOS-SERVER-XDZ

### DIFF
--- a/apps/synchronize/src/github/synchronizer.js
+++ b/apps/synchronize/src/github/synchronizer.js
@@ -69,12 +69,21 @@ export class GitHubSynchronizer {
     return { repositories, organizations };
   }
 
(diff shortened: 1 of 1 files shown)


gregberge (changes requested):
- We do not want to ignore all errors but just 404
- Code not tested (`return` without `return await` will not trigger `catch`)

jsfez:
> Reference in new issue

I can't write a test about this feature. That would require a Github App and an available installationId

jsfez:
> * We do not want to ignore all errors but just 404
> * Code not tested (`return` without `return await` will not trigger `catch`)

I can't write a test about it because it would require a real installation.

gregberge (approved):

PR #493, opened 2022-09-02
feat: detect default branch

jsfez:
fixes ARG-47


jsfez:
I'm confident with the app update but I have no clue how to test the build part.
Do you have any advice about it @gregberge ?

gregberge (approved):

gregberge:
> I'm confident with the app update but I have no clue how to test the build part.
> Do you have any advice about it @gregberge ?

Run a build in local and see what's going on.

apps/synchronize/src/github/synchronizer.js
```diff
@@ -117,23 +117,24 @@ export class GitHubSynchronizer {
...
-            baselineBranch: "master",
+            baselineBranch:
+              data.githubDefaultBranch !== repository.baselineBranch
+                ? repository.baselineBranch
+                : null,
           });
         }
 
-        return repository;
+        return Repository.query().insertAndFetch({ ...data, enabled: false });
```
gregberge, line 137: You replaced `insert` by `insertAndFetch`, why?

apps/graphql/src/definitions/Repository.js
```diff
@@ -170,21 +173,27 @@ export const resolvers = {
         return repository;
       });
     },
-    async updateBaselineBranch(source, { repositoryId, branchName }, context) {
+    async updateBaselineBranch(
+      source,
+      { repositoryId, baselineBranch = "" },
```
gregberge, line 178:
```suggestion
      { repositoryId, baselineBranch },
```
It is required, so I don't get the default, how can it be undefined?

apps/graphql/src/definitions/Repository.js
```diff
@@ -170,21 +173,27 @@ export const resolvers = {
...
+        const repository = await Repository.query(trx).findById(repositoryId);
         if (!repository) throw new APIError("Repository not found");
 
-        return repository;
+        const branchName = baselineBranch.trim();
+        return Repository.query(trx).patchAndFetchById(repositoryId, {
+          baselineBranch:
+            branchName && branchName !== repository.githubDefaultBranch
+              ? branchName
+              : "",
```
gregberge, line 195:
```suggestion
              : null,
```

apps/synchronize/src/github/synchronizer.js
```diff
@@ -117,23 +117,24 @@ export class GitHubSynchronizer {
...
         if (repository) {
-          await repository.$query().patchAndFetch(data);
-        } else {
-          repository = await Repository.query().insert({
+          return repository.$query().patchAndFetch({
             ...data,
-            enabled: false,
-            baselineBranch: "master",
+            baselineBranch:
+              data.githubDefaultBranch !== repository.baselineBranch
```
gregberge, line 131: So, if GitHub default branch is the same as baseline branch, we keep synchronizing it. Even if the user explicitely say I don't want to synchronize it. I think it is unclear.

For me it is simple.

- [ ] Use GitHub default branch

If checked, `baselineBranch` is `null`.

If unchecked, a field appear `baselineBranch` and it has to be filled by the user.

So let's removed the weird behaviour of testing if the branch is the same to reset the sync.

Also client-side let's be explicit with a checkbox.

gregberge (changes requested):

apps/synchronize/src/github/synchronizer.js
jsfez: I agree with your POV. I'll add the checkbox and remove the confusing behavior.

apps/synchronize/src/github/synchronizer.js
jsfez: I want to insure coherency between return cases.

Previously, this method returned not patched `repository` or `undefined`.

With my change, this method return the updated or created `repository`.

gregberge: But maybe if not need it should return undefined? Better to avoid a fetch for perf issue (but not very important).

apps/database/src/models/Repository.js
```diff
@@ -78,6 +80,12 @@ export class Repository extends Model {
     return this.constructor.getUsers(this.id, options);
   }
 
+  get referenceBranch() {
+    return !this.useDefaultBranch && this.baselineBranch
+      ? this.baselineBranch
+      : this.defaultBranch;
```
gregberge, line 86:
```suggestion
    return this.baselineBranch || this.defaultBranch
```

apps/app/src/components/Checkbox.js
```diff
@@ -0,0 +1,5 @@
+import * as React from "react";
+import { x } from "@xstyled/styled-components";
+import { Checkbox as AriakitCheckbox } from "ariakit/checkbox";
+
+export const Checkbox = (props) => <x.div as={AriakitCheckbox} {...props} />;
```
gregberge, line 5:
```suggestion
export const Checkbox = (props) => <x.checkbox as={AriakitCheckbox} {...props} />;
```

gregberge (changes requested):

PR #507, opened 2022-09-07
fix(build): avoid running diffs on same branch

gregberge:


jsfez (approved):

gregberge:
Not relevant, should be linked to another bug.

PR #531, opened 2022-09-18
docs(example): add GitHub Actions example + CircleCI example

gregberge:


jsfez (commented):
You should add a "screenshots" folder filled with two images in root folder and update CI commands accordingly

PR #541, opened 2022-09-19
feat(build): removed screenshots section

jsfez:
fixes ARG-119

Capture d’écran 2022-09-19 à 19 32 08


apps/app/src/pages/Build/ScreenshotsDiffCard.js
```diff
@@ -95,7 +95,7 @@ export function ScreenshotsDiffCard({
             as={ScreenshotDiffStatusIcon(screenshotDiff.status)}
             color={getStatusPrimaryColor(screenshotDiff.status)}
           />
-          {compareScreenshot.name || baseScreenshot.name}
+          {compareScreenshot?.name || baseScreenshot.name}
```
gregberge, line 98: Just leave base screenshot?

gregberge (approved):

apps/app/src/pages/Build/ScreenshotsDiffCard.js
jsfez: If we use only base screenshot name, there will be a bug on « added » Diff. This component is used in every screenshot sections

gregberge: Got it!

PR #546, opened 2022-09-22
feat(settings): split usage indicators

jsfez:
Standard Plan
Capture d’écran 2022-09-23 à 11 17 46

Free Plan
Capture d’écran 2022-09-23 à 11 17 32

No Plan
Capture d’écran 2022-09-23 à 11 02 52

No Plan with open details
Capture d’écran 2022-09-23 à 11 02 57

### DIFF
--- a/apps/app/src/pages/Owner/GeneralSettings.js
+++ b/apps/app/src/pages/Owner/GeneralSettings.js
@@ -1,27 +1,41 @@
 import * as React from "react";
 import { x } from "@xstyled/styled-components";
 import moment from "moment";
-import { CheckCircleFillIcon, LinkExternalIcon } from "@primer/octicons-react";
+import {
+  CheckCircleFillIcon,
+  ChevronRightIcon,
+  LinkExternalIcon,
+} from "@primer/octicons-react";
 import { gql } from "graphql-tag";
 import {
   Card,
   CardBody,
   CardHeader,
   CardTitle,
+  Disclosure,
+  DisclosureContent,
   Icon,
   Link,
+  LinkBlock,
   PrimaryTitle,
   ProgressBar,
   SidebarLayout,
   Time,
+  useDisclosureState,
 } from "@argos-ci/app/src/components";
 import config from "../../config";
-import { getPossessiveForm } from "../../modules/utils";
+import { getPossessiveForm, pluralize } from "../../modules/utils";
 
 export const OwnerSettingsFragment = gql`
   fragment OwnerSettingsFragment on Owner {
     name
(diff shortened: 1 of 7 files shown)


apps/app/src/pages/Owner/GeneralSettings.js
```diff
@@ -106,41 +119,143 @@ function PlanCard({ purchases, screenshotsLimitPerMonth, ...props }) {
...
     </Card>
   );
 }
 
 export function GeneralSettings({
-  owner: { name, purchases, currentMonthUsedScreenshots },
+  owner: { name, purchases, screenshotsLimitPerMonth, repositories },
 }) {
-  const screenshotsLimitPerMonth = Infinity;
+  const minScreenshotLimit = screenshotsLimitPerMonth || 7000;
```
gregberge, line 258: hmm ?

gregberge (changes requested):

apps/app/src/pages/Owner/GeneralSettings.js
jsfez: They are 2 interesting topic here :
- the ˋ|| 7000ˋ is here for people without plan purchase. By default repo have 7000 private screenshot / month even without plan. 

- the second topic is that I have here replaced `infinity` by a the plan's screenshots imite. It's maybe too early ?

gregberge: Actually it id unlimited for every one, so please remove this fake limit.

apps/app/src/pages/Owner/GeneralSettings.js
```diff
@@ -42,7 +56,7 @@ const FeatureIcon = (props) => <Icon mt="-1px" {...props} />;
 const PlanName = (props) => <x.div fontSize="4xl" mb={3} {...props} />;
 
 const Plan = ({ purchase }) => {
-  if (!purchase.plan) return <PlanName mb={2}>No plan</PlanName>;
+  if (!purchase?.plan) return <PlanName mb={2}>No plan</PlanName>;
```
gregberge, line 59: Does not exist, we fallback on Free plan server-side.

gregberge (changes requested):

apps/app/src/pages/Owner/GeneralSettings.js
gregberge, line 59: No plan does not exist. It fallbacks on free plan. If not plan is defined in database then we have a special case where we say that there is no plan configured on this Argos instance.

gregberge (changes requested):

PR #547, opened 2022-09-23
feat(build): add plan limit banner

jsfez:
Resolves ARG-52 

- feat(build): add overcapacity banner
- feat(build): disable review button for overcapacity private repo

Capture d’écran 2022-09-23 à 10 04 26
Capture d’écran 2022-09-23 à 10 04 52
Capture d’écran 2022-09-23 à 10 05 48
Capture d’écran 2022-09-23 à 10 05 55

### DIFF
--- a/apps/app/src/pages/Build/ReviewButton.js
+++ b/apps/app/src/pages/Build/ReviewButton.js
@@ -4,13 +4,13 @@ import {
   Alert,
   Button,
   Menu,
-  MenuButton,
   MenuItem,
   useMenuState,
   MenuButtonArrow,
   useTooltipState,
   TooltipAnchor,
   Tooltip,
+  HeadlessMenuButton,
 } from "@argos-ci/app/src/components";
 import { gql, useMutation } from "@apollo/client";
 import { StatusIcon } from "../../containers/Status";
@@ -63,7 +63,7 @@ export function ReviewButtonContent({ repository, disabled }) {
 
   return (
     <x.div display="flex" flexDirection="column" flex={1}>
-      <MenuButton
+      <HeadlessMenuButton
         as={Button}
         state={menu}
         disabled={disabled || loading}
@@ -71,7 +71,7 @@ export function ReviewButtonContent({ repository, disabled }) {
       >
         Review changes
         <MenuButtonArrow state={menu} />
-      </MenuButton>
+      </HeadlessMenuButton>
       <Menu aria-label="Review changes" state={menu}>
         <MenuItem
           state={menu}
--- a/apps/app/src/pages/Build/ReviewButton.js
+++ b/apps/app/src/pages/Build/ReviewButton.js
@@ -8,6 +8,9 @@ import {
   MenuItem,
   useMenuState,
   MenuButtonArrow,
+  useTooltipState,
+  TooltipAnchor,
+  Tooltip,
 } from "@argos-ci/app/src/components";
 import { gql, useMutation } from "@apollo/client";
 import { StatusIcon } from "../../containers/Status";
@@ -24,10 +27,17 @@ export const ReviewButtonBuildFragment = gql`
 export const ReviewButtonRepositoryFragment = gql`
   fragment ReviewButtonRepositoryFragment on Repository {
     permissions
+    private
   }
 `;
 
-export function ReviewButtonContent({ repository }) {
+export const ReviewButtonOwnerFragment = gql`
+  fragment ReviewButtonOwnerFragment on Owner {
+    consumptionRatio
+  }
+`;
+
+export function ReviewButtonContent({ repository, disabled }) {
   const { id, status } = repository.build;
   const menu = useMenuState({ placement: "bottom-end", gutter: 4 });
   const [setValidationStatus, { loading, error }] = useMutation(gql`
@@ -53,10 +63,15 @@ export function ReviewButtonContent({ repository }) {
 
   return (
     <x.div display="flex" flexDirection="column" flex={1}>
-      <Button as={MenuButton} state={menu} alignSelf="end" disabled={loading}>
+      <MenuButton
+        as={Button}
+        state={menu}
+        disabled={disabled || loading}
+        alignSelf="end"
+      >
         Review changes
         <MenuButtonArrow state={menu} />
-      </Button>
+      </MenuButton>
       <Menu aria-label="Review changes" state={menu}>
         <MenuItem
           state={menu}
@@ -99,10 +114,32 @@ export function ReviewButtonContent({ repository }) {
   );
 }
 
+function DisabledReviewButton({ repository }) {
+  const tooltip = useTooltipState();
+
+  return (
+    <>
+      <TooltipAnchor state={tooltip}>
+        <ReviewButtonContent repository={repository} disabled />
+      </TooltipAnchor>
+      <Tooltip state={tooltip}>
+        You have hit 100% of your screenshots usage. Please upgrade to unlock
+        build reviews.
+      </Tooltip>
+    </>
+  );
+}
+
 export function ReviewButton({ repository }) {
-  return ["accepted", "rejected", "diffDetected"].includes(
-    repository.build.status
-  ) ? (
-    <ReviewButtonContent repository={repository} />
-  ) : null;
+  if (
+    !["accepted", "rejected", "diffDetected"].includes(repository.build.status)
+  ) {
+    return null;
+  }
+
+  if (repository.private && repository.owner.consumptionRatio >= 1) {
+    return <DisabledReviewButton repository={repository} />;
+  }
+
+  return <ReviewButtonContent repository={repository} />;
 }
--- a/apps/app/src/pages/Build/index.js
+++ b/apps/app/src/pages/Build/index.js
@@ -1,3 +1,4 @@
+/* eslint-disable react/no-unescaped-entities */
 import * as React from "react";
 import { gql } from "graphql-tag";
 import { useParams } from "react-router-dom";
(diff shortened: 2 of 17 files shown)


apps/graphql/src/definitions/Owner.js
```diff
@@ -151,6 +153,27 @@ export const resolvers = {
...
+        .where("screenshotBucket:repository.private", "true")
+        .where(
+          "screenshots.createdAt",
+          ">=",
+          moment().startOf("month").toISOString()
+        )
+        .resultSize();
+      const purchases = await getActivePurchases(owner);
+      if (purchases.length === 0) {
+        return 0;
```
gregberge, line 169: If there is no purchase, we fallback to the free plan. That is currently unlimited, but later not. So no "0" in the code, but read it from database.

apps/graphql/src/definitions/Owner.js
```diff
@@ -151,6 +153,27 @@ export const resolvers = {
...
+      }
+
+      const plan = await purchases[0].$relatedQuery("plan");
```
gregberge, line 172: Use `withGraphFetched` on purchases.

apps/app/src/pages/Build/ReviewButton.js
```diff
@@ -99,10 +109,29 @@ export function ReviewButtonContent({ repository }) {
   );
 }
 
+function DisabledReviewButton() {
+  const tooltip = useTooltipState();
+
+  return (
+    <>
+      <TooltipAnchor state={tooltip}>
+        <Button disabled>Review changes</Button>
```
gregberge, line 118: Use the same button, wording is duplicated.

gregberge (changes requested):
+ protect the mutation server side to avoid reviewing

apps/app/src/pages/Build/index.js
```diff
@@ -128,6 +148,30 @@ function BuildStat({ status, count, label }) {
...
+  return (
+    <Alert
+      severity={owner.consumptionRatio >= 1 ? "danger" : "warning"}
+      mt={-3}
+      mb={3}
+      w="fit-content"
+      mx="auto"
+    >
+      You've hit {consumptionThreshold} of the{" "}
+      {owner.purchases.plan?.name || "Free"} plan limit.{" "}
```
gregberge, line 169: purchases should be an array. Also to get the active plan we should not go from purchase. The fallback to free plan must be done server-side.

gregberge (changes requested):
Also Argos can be self-hosted and installed locally. In this case there is no plan in database, so the plan should not appear in this case. We have to handle it.

jsfez:
This PR can't be merged because it assumes that each repository is linked to an account. For now, here are a lot of missing accounts in prod database because only the purchases event create an account.

We have to create missing account in PROD database and automate account creation during repositories synchronization. Before merging this PR

apps/app/src/pages/Build/ReviewButton.js
```diff
@@ -54,7 +66,7 @@ export function ReviewButtonContent({ repository }) {
   return (
     <x.div display="flex" flexDirection="column" flex={1}>
       <Button as={MenuButton} state={menu} alignSelf="end" disabled={loading}>
```
gregberge, line 68: It is the opposite `<MenuButton as={Button}`>

apps/app/src/pages/Build/ReviewButton.js
```diff
@@ -54,7 +66,7 @@ export function ReviewButtonContent({ repository }) {
...
-        Review changes
+        {BUTTON_LABEL}
```
gregberge, line 69: Why not creating a component instead of just sharing the label?

apps/app/src/pages/Build/ReviewButton.js
```diff
@@ -99,10 +111,32 @@ export function ReviewButtonContent({ repository }) {
...
+function DisabledReviewButton() {
+  const tooltip = useTooltipState();
+
+  return (
+    <>
+      <TooltipAnchor state={tooltip}>
+        <Button disabled>{BUTTON_LABEL}</Button>
+      </TooltipAnchor>
+      <Tooltip state={tooltip}>
+        You have it 100% of your screenshots usage. Please upgrade to unlock
```
gregberge, line 123:
```suggestion
        You have hit 100% of your screenshots usage. Please upgrade to unlock
```

apps/database/src/models/Plan.js
```diff
@@ -30,7 +30,6 @@ export class Plan extends Model {
     const freePlan = await Plan.query()
       .whereRaw(`"name" ILIKE '%free%'`)
```
gregberge, line 31: Use equality instead of ILIKE.

apps/graphql/src/definitions/Build.js
```diff
@@ -200,6 +200,19 @@ export const resolvers = {
...
 
+      const build = await Build.query()
+        .findById(buildId)
+        .withGraphFetched("repository");
+
+      if (build.repository.private) {
+        const account = Account.getAccount(build.repository);
+        const screenshotsConsumptionRatio =
+          await account.getScreenshotsConsumptionRatio();
+        if (screenshotsConsumptionRatio >= 1.1) {
```
gregberge, line 211: This limit is used several times, put it in a constant please.

gregberge (changes requested):

gregberge (changes requested):
It could maybe no have account.

apps/app/src/pages/Build/ReviewButton.js
```diff
@@ -24,9 +27,23 @@ export const ReviewButtonBuildFragment = gql`
...
 `;
 
+export const ReviewButtonOwnerFragment = gql`
+  fragment ReviewButtonOwnerFragment on Owner {
+    consumptionRatio
+  }
+`;
+
+const ReviewChangesButton = React.forwardRef(({ state, ...props }, ref) => (
+  <Button alignSelf="end" ref={ref} {...props}>
```
gregberge, line 41: ref first please.

apps/app/src/pages/Build/ReviewButton.js
```diff
@@ -24,9 +27,23 @@ export const ReviewButtonBuildFragment = gql`
...
+    Review changes
+    <MenuButtonArrow state={state} />
```
gregberge, line 43: Without state it is not correct. You have to pass this in children.

apps/app/src/pages/Build/ReviewButton.js
```diff
@@ -99,10 +113,32 @@ export function ReviewButtonContent({ repository }) {
   );
 }
 
+function DisabledReviewButton() {
+  const tooltip = useTooltipState();
+
+  return (
+    <>
+      <TooltipAnchor state={tooltip}>
+        <ReviewChangesButton disabled />
```
gregberge, line 122: So this usage is invalid, since you don't pass `state`.

gregberge: You are struggling because the whole Button + Menu should be in the same component, and it is that one that you disable.

apps/graphql/src/definitions/Build.js
```diff
@@ -200,6 +200,24 @@ export const resolvers = {
...
+        if (account) {
+          const screenshotsConsumptionRatio =
+            await account.getScreenshotsConsumptionRatio();
+          if (
+            screenshotsConsumptionRatio &&
+            screenshotsConsumptionRatio >= 1.1
+          ) {
+            throw new APIError("Please upgrade to unlock build reviews.");
+          }
+        }
```
gregberge, line 218: Without account, we should consider it free. Adding `if (account)` everywhere does not solve it. You have to go with the way of `Account.getAccount` always returning an account, even if this account does not really exist in database.

apps/web/src/middlewares/repoAuth.js
```diff
@@ -18,16 +18,19 @@ export const repoAuth = [
...
-          402,
-          `Build refused for insufficient credit. Thank to upgrade Argos plan`
-        );
+        if (screenshotsConsumptionRatio && screenshotsConsumptionRatio >= 1.1) {
+          throw new HttpError(
+            402,
+            `Build refused for insufficient credit. Thank to upgrade Argos plan`
+          );
+        }
       }
```
gregberge, line 34: This code should not be part of `repoAuth`, this is not the same responsibility.

gregberge (changes requested):

apps/graphql/src/definitions/Build.js
jsfez: I'm ok to use a constant but I don't know where to store it. Until I find a good place, I'll stick to hard coded values

apps/app/src/pages/Build/ReviewButton.js
jsfez: As I lose too much time with this "use the same button that is not a button" game, I will rollback to two different components. The main purpose was to no duplicate the button label, so I'll use a common variable label.

apps/app/src/pages/Build/ReviewButton.js
jsfez: no more relevant.

apps/app/src/pages/Build/ReviewButton.js
jsfez: no more relevant.

apps/web/src/middlewares/repoAuth.js
jsfez: I'll be removed from the `repoAuth` repo then. I'm sure that our users will be happy too!

apps/app/src/pages/Build/ReviewButton.js
gregberge: Ok I will do it then.

apps/database/src/models/Account.js
```diff
@@ -117,7 +119,30 @@ export class Account extends Model {
...
+        .findOne("userId", userId)
+        .withGraphFetched("user");
+      return userAccount || null;
+    }
+
+    const organizationAccount = await Account.query()
+      .findOne("organizationId", organizationId)
+      .withGraphFetched("organization")
+      .first();
+    return organizationAccount || null;
```
gregberge, line 146:
```suggestion
    if (userId) {
      const userAccount = await Account.query()
        .findOne("userId", userId)
        .withGraphFetched("user");
      return userAccount || Account.fromJSON({ userId });
    }
    
    if (organizationId) {
    const organizationAccount = await Account.query()
      .findOne("organizationId", organizationId)
      .withGraphFetched("organization")
      .first();
    return organizationAccount || Account.fromJSON({ organizationId });
    }
    
    throw new Error("Can't get account without userId or organizationId")
```

gregberge (changes requested):

PR #549, opened 2022-09-23
feat(api): block overcapacity upload

jsfez:


apps/database/src/models/Account.js
```diff
@@ -37,4 +37,8 @@ export class Account extends Model {
       },
     };
   }
+
+  get type() {
+    return this.userId ? "user" : "organization";
+  }
```
gregberge: `virtualAttributes` not needed for that?

gregberge: Also we should do:

```js
if (this.userId && this.accountId) throw new Error(`Invariant incoherent account type`)
if (this.userId) return "user"
if (this.accountId) return "organization"
throw new Error(`Invariant incoherent account type`)

File: apps/web/src/middlewares/repoAuth.js [hunk 2]
```diff
@@ -1,8 +1,54 @@
...
+  const date = new Date();
+  date.setDate(1);
+  return new Date(date.toISOString().split("T")[0]);
+}
+
+async function getScreenshotConsumption({ userId, organizationId }) {
+  const query = Screenshot.query()
+    .joinRelated("screenshotBucket.repository")
+    .where("screenshotBucket:repository.private", "true")
+    .where("screenshots.createdAt", ">=", getStartOfMonth());
```
gregberge, line 21: Use SQL for that, don't rely on JS.

apps/web/src/middlewares/repoAuth.js
```diff
@@ -1,8 +1,54 @@
...
+  const account = await Account.query()
+    .where("userId", repository.userId)
+    .orWhere("organizationId", repository.organizationId)
+    .first();
+
+  const purchase = await Purchase.query()
+    .withGraphFetched("plan")
+    .where("accountId", account.id)
+    .where((query) =>
+      query.whereNull("endDate").orWhere("endDate", ">=", new Date())
```
gregberge, line 42: same, use SQL.

apps/database/src/models/Account.js
```diff
@@ -37,4 +40,69 @@ export class Account extends Model {
...
+      .withGraphFetched("plan")
+      .first();
+
+    return purchase || null;
+  }
+
+  async getScreenshotsLimitPerMonth() {
+    const purchase = await this.getActivePurchase();
+    if (purchase) {
+      return purchase.plan.screenshotsLimitPerMonth === -1
```
gregberge, line 60: Factorize this logic on the plan.

apps/database/src/models/Account.js
```diff
@@ -37,4 +40,69 @@ export class Account extends Model {
...
+    const purchase = await Purchase.query()
+      .where("accountId", this.id)
+      .where("startDate", "<", "now()")
+      .where((query) =>
+        query.whereNull("endDate").orWhere("endDate", ">=", "now()")
+      )
+      .withGraphFetched("plan")
+      .first();
+
+    return purchase || null;
```
gregberge, line 54: ??

gregberge (approved):
Nothing critical.

apps/database/src/models/Account.js
jsfez: - You: ??
- The server: 🔥

PR #550, opened 2022-09-23
fix(build): emtpy build detection

jsfez:


gregberge (approved):

gregberge (changes requested):
The code is hard to maintain because components are not correctly isolated. 

The build should be sufficient to got the informations. Here the problem is you have to pass several objects and the lack of fragment. More the product will grow more it will be difficult.

PR #551, opened 2022-09-24
feat(marketplace-event): rework event handle

jsfez:


apps/synchronize/src/github/eventHelpers.js
```diff
@@ -0,0 +1,78 @@
...
+      .findOne("user.githubId", githubId);
+    return account ?? null;
+  }
+  if (type.toLowerCase() === "organization") {
+    const account = await Account.query()
+      .joinRelated("organization")
+      .findOne("organization.githubId", githubId);
+    return account ?? null;
+  }
+  if (!type) throw new Error(`can't find account of type : ${type}`);
```
salper, line 20: I think this test makes no sense. You can just throw the error here.

apps/database/src/models/Account.js
```diff
@@ -35,6 +36,40 @@ export class Account extends Model {
...
+  async getActivePurchase() {
+    const today = new Date().toISOString();
+    return Purchase.query()
+      .where("accountId", this.id)
+      .where("startDate", "<", today)
+      .where((query) =>
+        query.whereNull("endDate").orWhere("endDate", ">=", today)
+      )
+      .joinRelated("plan")
+      .orderBy("endDate")
```
salper, line 72: Can you have more than one active plan?

apps/synchronize/src/github/eventHelpers.js
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export async function getAccountOrThrow(payload) {
+  const account = await getAccount(payload);
```
salper, line 24: You can probably do an early return here:
```js
if (account) {
  return account;
}
```

apps/synchronize/src/github/eventHerpers.test.js
```diff
@@ -0,0 +1,230 @@
...
+  getActivePurchaseOrThrow,
+  getNewPlanOrThrow,
+} = require("./eventHelpers");
+
+describe("event helpers", () => {
+  useDatabase();
+
+  describe("#getAccount", () => {
+    it("of missing user should return null", async () => {
+      const account = await getAccount(USER_PURCHASE_EVENT_PAYLOAD);
```
salper, line 21: Not really mind blowing, but you can also do:
```js
await expect(getAccount(USER_PURCHASE_EVENT_PAYLOAD)).resolves.toBeNull()
```

apps/synchronize/src/github/eventHerpers.test.js
```diff
@@ -0,0 +1,230 @@
...
+        organizationId: organization.id,
+      });
+      account = await getOrCreateAccount(ORGANIZATION_PURCHASE_EVENT_PAYLOAD);
+      expect(account).toMatchObject({ organizationId: organization.id });
+    });
+  });
+
+  describe("#getActivePurchaseOrThrow", () => {
+    it("should throw when account is not provided", async () => {
+      expect.assertions(1);
```
salper, line 165: You can also do:
```js
await expect(getActivePurchaseOrThrow()).rejects.toMatch('can't find purchase of missing account')
```

salper (commented):
Just passing by... 😁

jsfez (commented):
Thank you for your review !

apps/synchronize/src/github/eventHelpers.js
jsfez: yes

gregberge (changes requested):
`eventHerpers` -> `eventHelpers`

apps/synchronize/src/github/events.js
```diff
@@ -14,20 +17,35 @@ export async function handleGitHubEvents({ name, payload }) {
...
-          case "pending_change": {
-            await change(payload);
+          case "changed": {
+            const newPlan = await getNewPlanOrThrow(payload);
+            const account = await getAccountOrThrow(payload);
+            const purchase = await getActivePurchaseOrThrow(account);
+            await Purchase.query()
+              .patch({ endDate: payload.effective_date })
+              .findById(purchase.id);
+            await Purchase.query().insertAndFetch({
```
gregberge, line 36: Transaction + `Promise.all`

gregberge: Why `insertAndFetch` and not just `insert` ?

apps/synchronize/src/github/events.js
```diff
@@ -14,20 +17,35 @@ export async function handleGitHubEvents({ name, payload }) {
...
         switch (payload.action) {
           case "purchased": {
-            await purchase(payload);
+            const plan = await getNewPlanOrThrow(payload);
+            const account = await getOrCreateAccount(payload);
+            Purchase.query().insertAndFetch({
+              accountId: account.id,
+              planId: plan.id,
+              startDate: payload.effective_date,
+            });
```
gregberge, line 26: `insert`, not `insertAndFetch`.

gregberge (changes requested):
We should have a method to ensure that there is only one active plan whatever. This method will check all is coherent at the end of each event.

These events sent by GitHub are volatile. It means if our app is unresponsive when they send it, the app will not be in sync. It is a problem because it could result of buying a plan is not effective in Argos and there is no solution except contacting us.

To avoid this, we should synchronise plans of the user / owner in our Synchronizer. This way, when the use log in we are sure that all is in sync.

https://docs.github.com/en/developers/github-marketplace/using-the-github-marketplace-api-in-your-app/rest-endpoints-for-the-github-marketplace-api

apps/synchronize/src/github/eventHerpers.test.js
gregberge: Since I never remember the syntax, I use a simple try / catch + `expect.assertions`. But it is tricky because if you remove the first line your test become irrelevant and the two lines are not linked and not statically checkable.

Your version is way better @salper.

PR #555, opened 2022-09-27
feat: synchronize purchases

jsfez:


apps/synchronize/src/github/synchronizer.js
```diff
@@ -135,6 +138,40 @@ export class GitHubSynchronizer {
...
   }
 
+  async synchronizePurchase(accountId) {
+    try {
+      const { data: payload } =
+        await this.octokit.apps.getSubscriptionPlanForAccount({
+          account_id: accountId,
+        });
+      const newPlan = await getNewPlanOrThrow(payload);
+      const account = await getOrCreateAccount(payload);
```
gregberge, line 148: Promise.all

apps/synchronize/src/github/synchronizer.js
```diff
@@ -135,6 +138,40 @@ export class GitHubSynchronizer {
...
+            accountId: account.id,
+            planId: newPlan.id,
+            startDate: payload.effective_date,
+          }),
+        ]);
+      });
+    } catch (error) {
+      if (error.status === 404) {
+        return;
+      }
```
gregberge, line 169: A 404 is no plan, so it is equivalent to "cancelled".

apps/synchronize/src/github/synchronizer.js
```diff
@@ -135,6 +138,40 @@ export class GitHubSynchronizer {
...
+        await Promise.all(() => [
+          Purchase.query(trx)
+            .patch({ endDate: payload.effective_date })
+            .findById(activePurchase.id),
+          Purchase.query(trx).insert({
+            accountId: account.id,
+            planId: newPlan.id,
+            startDate: payload.effective_date,
+          }),
+        ]);
```
gregberge, line 164: The code is duplicated. The logic is the same as "purchased" and "cancelled", so the code should be refactored and used both in events and in sync.

gregberge (changes requested):

PR #562, opened 2022-10-05
webdriverIO example

jsfez:
- doc(examples): udpate gitignore files
- doc(example): add webdriverIO example


gregberge (approved):

gregberge (changes requested):
Use kebab case for example name. Else it is not coherent with circleci and I think it is better to stick to kebab for readability.

PR #567, opened 2022-10-07
add diff transparency

jsfez:


gregberge (approved):

gregberge:
@jsfez do not merge now, else the visibility will not be optimal in the UI

PR #570, opened 2022-10-10
fix(build): prevent missing build type

jsfez:
fixes ARG-171


gregberge (changes requested):
At this point we don't know the type. That's why it is null. The fix is not to fill a wrong type but to handle "not yet a type".

gregberge (changes requested):
I think the build diffs are created when the build is processed. So before the build type is empty, the GraphQL is not correct and neither the front-end.

We can't know the build type before processing it. This refactor does not solve the problem at all.

gregberge (approved):

PR #590, opened 2022-10-17
rework button

jsfez:
- feat: add color modes
- feat: remove 'background' color
- feat: rework button


apps/app/src/components/Menu.js
```diff
@@ -67,7 +67,7 @@ export const MenuButtonArrow = ({ as, ...props }) => {
 export const Menu = (props) => (
   <x.menu
     as={HeadlessMenu}
-    backgroundColor="background"
+    backgroundColor="lighter"
```
gregberge, line 70: `lighter` should not be used directly, please use `bg` alias instead.

gregberge (changes requested):

PR #593, opened 2022-10-18
feat: add IconButton component

jsfez:


stories/IconButton.stories.js
```diff
@@ -0,0 +1,32 @@
...
+import { EyeIcon } from "@heroicons/react/24/solid";
+import { IconButton } from "@argos-ci/app/src/components";
+
+const main = {
+  title: "Icon Button",
+};
+
+export default main;
+
+export const Primary = () => (
```
gregberge, line 13: ??

gregberge (changes requested):

stories/IconButton.stories.js
jsfez: I can't put a `??` here.

gregberge: 😂

PR #602, opened 2022-10-25
feat: load inter font

jsfez:


apps/app/src/components/GlobalStyle.js
```diff
@@ -19,6 +20,7 @@ export const GlobalStyle = createGlobalStyle`
     height: 100%;
     background-color: bg;
     color: primary-text;
+    font-family: default;
```
gregberge, line 23: Nop

gregberge (changes requested):

apps/app/src/components/GlobalStyle.js
```diff
@@ -71,3 +73,10 @@ export const GlobalStyle = createGlobalStyle`
     outline: 5px auto -webkit-focus-ring-color;
   }
 `;
+
+export const GlobalStyle = () => (
+  <>
+    <ArgosGlobalStyle />
+    <Preflight />
```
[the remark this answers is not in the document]
jsfez: OK. FYI, i've get the code from WWW

apps/app/src/components/GlobalStyle.js
jsfez: This line alone load the Inter as the default font. If I remove it, this PR is useless

PR #613, opened 2022-10-28
feat(new-build): connect review changes button

jsfez:


apps/app/src/utils/externalLinks.js
```diff
@@ -0,0 +1,7 @@
+import config from "../config";
+
+export default {
+  githubLogin: `${config.get("github.loginUrl")}&redirect_uri=${
+    window.location.origin
+  }/auth/github/callback?r=${encodeURIComponent(window.location.pathname)}`,
```
gregberge, line 6: Replace it by a function `window.location.pathname` is not a constant.

gregberge (changes requested):

PR #619, opened 2022-10-29
various improvements

gregberge:
- chore: remove import * as React
- chore(prettier): sort imports


jsfez (approved):

jsfez:
Gros problème sur le diff Argo !

PR #625, opened 2022-11-01
feat: use TypeScript + ESM

gregberge:
Also move on JWT instead of using GitHub token to login.


apps/graphql/src/definitions/Build.ts
```diff
@@ -228,10 +238,11 @@ export const resolvers = {
           `)
         )
         .count("*")
-        .groupBy(1);
+        .groupBy("1");
```
jsfez, line 241:
```suggestion
        .groupBy("status");
```

PR #626, opened 2022-11-02
feat(new-build): hotkey dialog and various fixes

jsfez:


apps/app/src/pages/Build/Hotkeys.js
```diff
@@ -0,0 +1,30 @@
+export const HOTKEYS = {
+  toggleChangesOverlay: {
+    shortcut: "H",
+    description: "Show/Hide changes overlay",
```
gregberge, line 4: Make it dynamic.

gregberge (changes requested):
Description should be what is displayed in tooltip (dynamic) and also a static version for dialog.

apps/app/src/pages/Build/Hotkeys.js
jsfez: I like it like that

PR #628, opened 2022-11-04
feat: store file size

jsfez:


apps/build/src/createBuildDiffs.ts
```diff
@@ -59,11 +59,19 @@ const getJobStatus = ({
...
   sameFileId,
+  compareScreenshot,
 }: {
   compareWithBaseline: boolean;
-  baseScreenshot: boolean;
+  baseScreenshot: Screenshot | null;
   sameFileId: boolean;
+  compareScreenshot: Screenshot;
 }) => {
+  if (baseScreenshot && baseScreenshot.file && !baseScreenshot.file.height) {
```
gregberge, line 69:
```suggestion
  if (baseScreenshot && baseScreenshot.file?.height) {
```

apps/build/src/createBuildDiffs.ts
```diff
@@ -124,8 +132,9 @@ export const createBuildDiffs = async (build: Build) => {
           compareScreenshotId: compareScreenshot.id,
           jobStatus: getJobStatus({
             compareWithBaseline,
-            baseScreenshot: Boolean(baseScreenshot),
+            baseScreenshot: baseScreenshot || null,
```
gregberge, line 135:
```suggestion
            baseScreenshot: baseScreenshot ?? null,
```

apps/screenshot-diff/src/computeScreenshotDiff.ts
```diff
@@ -40,12 +41,16 @@ export const computeScreenshotDiff = async (
   const diffResultPath = join(tmpDir, "diff.png");
 
   await Promise.all([
-    download({
-      s3,
-      outputPath: baseScreenshotPath,
-      Bucket: bucket,
-      Key: screenshotDiff.baseScreenshot!.s3Id,
-    }),
+    ...(screenshotDiff.baseScreenshot
```
gregberge, line 44: This code is not well factored. It has to be rewritten. If we have several files that can be downloaded or not, then we have to manage them differently to be able to unlink them properly. We have to create an abstraction on the S3 file.

```js
const baseScreenshot = new S3Image({ bucket, key })

await baseScreenshot.download()
// Will failed if a `download` has not be done before
const baseScreenshotFilepath = baseScreenshot.getFilePath()
// Will unlink or do nothing depending if a dowload has been made or not
await baseScreenshot.unlink()
// Get image dimensions, cached
await baseScreenshot.getDimensions()
```

In the `imageDifference` we will use these file too, to benefit of cached of `getDimensions`.

apps/screenshot-diff/src/computeScreenshotDiff.ts
```diff
@@ -54,19 +59,59 @@ export const computeScreenshotDiff = async (
...
+      .filter(
+        (screenshot) =>
+          screenshot &&
+          screenshot.fileId &&
+          screenshot.file &&
+          (screenshot.file.width === undefined ||
+            screenshot.file.height === undefined)
+      )
+      .map(async (screenshot) => {
+        const { width, height } = await getImageSize(screenshot.filePath);
```
gregberge, line 76: `getImageSize` is already done in the diff compute, so here we will compute the image size twice. We should memoize it to avoid computing size multiple times.

apps/screenshot-diff/src/computeScreenshotDiff.ts
```diff
@@ -54,19 +59,59 @@ export const computeScreenshotDiff = async (
...
+  }
+
   const difference = await diffImages({
     actualFilename: compareScreenshotPath,
     expectedFilename: baseScreenshotPath,
     diffFilename: diffResultPath,
   });
 
-  let uploadResult = null;
+  let file = null;
```
gregberge, line 101:
```suggestion
  let diffFile = null;
```

apps/screenshot-diff/src/computeScreenshotDiff.ts
```diff
@@ -54,19 +59,59 @@ export const computeScreenshotDiff = async (
...
+  if (
+    !screenshotDiff.baseScreenshot ||
+    (screenshotDiff.compareScreenshot!.fileId !== null &&
+      screenshotDiff.baseScreenshot!.fileId ===
+        screenshotDiff.compareScreenshot!.fileId)
+  ) {
+    await ScreenshotDiff.query().findById(screenshotDiff.id).patch({
+      jobStatus: "complete",
+    });
+    return;
```
gregberge, line 92: So here, no unlink at all...

apps/screenshot-diff/src/computeScreenshotDiff.ts
```diff
@@ -54,19 +59,59 @@ export const computeScreenshotDiff = async (
...
-  let uploadResult = null;
+  let file = null;
   if (difference.score > 0) {
-    uploadResult = await upload({
+    const uploadResult = await upload({
       s3,
       Bucket: bucket,
       inputPath: diffResultPath,
     });
+    file = await File.query()
```
gregberge, line 108: We don't have any transaction, so if the code failed after adding the file, we will have twice the same file in database. Since file key is unique it will break and the diff will never be processed.

gregberge: We have to insert file + screenshot diff in a transaction.

apps/screenshot-diff/src/util/image-diff/imageDifference.ts
```diff
@@ -36,7 +36,7 @@ const transparent = (
   return gmImage;
 };
 
-const getImageSize = async (filename: string) => {
+export const getImageSize = async (filename: string) => {
```
gregberge, line 39: What is the relation between `imageDifference.ts` and `getImageSize` ?

gregberge (changes requested):

apps/screenshot-diff/src/computeScreenshotDiff.ts
jsfez: When you say that getFilePath should fail, do you mean throw an error or return `undefined` ?

gregberge: Throw because it should not be called if download is not called before, it would mean the code is incoherent.

PR #639, opened 2022-11-14
fix: fix screenshot diff status

jsfez:
We forgot to patch the score when there are only negligible diffs. This mistake create a new type of diff with filled baseScreenshotId, filled compareScreenshotId but without score.

First, I have fixed our mistake. Then, I have updated the screenshot diff status algo: 
- No compareScreenshotId => "removed"
- No baseScreenshotId + "failed" in name => "failed"
- No baseScreenshotId => "added"
- score && score > 0 => "updated"
- else => "stable"


gregberge (changes requested):
Can you apply exactly the same algorithm in sorting?

I mean you can do:

```sql
case ... 
  then 10;  -- removed
case ...
  then 5; -- failed
```

The same order with a comment indicating which is a state and a score for each state? Thanks!

PR #653, opened 2022-11-21
feat: add empty build message

jsfez:


jsfez:
It's not at the same place than other because I want to display the sidebar.

As you have removed the getStatusLabel function, I thought that I should duplicate the message. 

Capture d’écran 2022-11-21 à 15 03 00
Capture d’écran 2022-11-21 à 15 03 09

PR #668, opened 2022-11-26
feat: redesign header of the page

gregberge:
Remove footer.


jsfez (approved):

jsfez:
I would prefer to keep the max-width because the amount of displayed data is small.

PR #683, opened 2022-12-10
remove s3 image postfix

jsfez:
- fix: remove image duplicate extension
- feat: set score to 1 when diff detected


apps/screenshot-diff/src/util/image-diff/index.ts
```diff
@@ -23,7 +23,7 @@ export const diffImages = async ({
   });
 
   return {
-    score: difference.value,
+    score: difference.value < 0.00005 ? 0 : difference.value,
```
gregberge, line 26: Why? Thresold should be sufficient. For big images 1% is enormous, for a small one it is just a pixel. So it makes no sense to have a check here.

apps/screenshot-diff/src/util/image-diff/test/e2e.test.ts
```diff
@@ -122,4 +122,16 @@ describe("e2e", () => {
 
     expect(difference).toMatchSnapshot();
   });
+
+  it("find differences on similar images", async () => {
```
gregberge, line 126: completely different you mean? If the result value is 1 ?

apps/storage/src/S3Image.ts
```diff
@@ -87,9 +87,7 @@ export class S3ImageFile extends AbstractImageFile implements ImageFile {
     if (!this.key) {
       throw new Error("Missing key");
     }
-    const outputPath = await tmpName({
-      postfix: ".jpg",
```
gregberge, line 91: 🙄

gregberge (changes requested):

apps/screenshot-diff/src/util/image-diff/index.ts
jsfez: A agree that percentage is not the best indicator. Nonetheless, the [border test](apps/screenshot-diff/src/util/image-diff/index.test.ts) return a score around "0.00001272". Before odiff, we ignored this kind of diff. What should we do now?

apps/screenshot-diff/src/util/image-diff/test/e2e.test.ts
jsfez: Similar for humans and for odiff. But you are wright, I can find a better name

apps/screenshot-diff/src/util/image-diff/index.ts
gregberge: OK

gregberge (approved):

PR #685, opened 2022-12-10
feat: diff in reference builds

jsfez:


gregberge (changes requested):
The build notification should not be "failure" if the build is a reference one. Always.

gregberge (approved):

PR #692, opened 2022-12-28
feat: add missing accounts

jsfez:


apps/web/src/api/auth.ts
```diff
@@ -44,6 +44,7 @@ async function registerUserFromGitHub(accessToken: string) {
     await User.query().findById(user.id).patch(userData);
   } else {
     user = await User.query().insertAndFetch(userData);
+    await Account.query().insert({ userId: user.id });
```
gregberge, line 47: And the organization? In synchronizer it is not done.

gregberge (changes requested):

apps/web/src/api/auth.ts
jsfez: First, there is no organization without account in the database. Moreover, I have check the code and I didn't find any line where an Organization is created without account. I think we already take care of creating account after Organization creation.

gregberge (approved):

PR #715, opened 2023-01-05
pr number info

jsfez:
- feat: add PR number to build info
- feat: branch and commit on same info line
- feat: enlarge commit column on build list


gregberge (changes requested):
I don't think it is a good thing to group commit and branch.

gregberge (approved):

PR #733, opened 2023-02-19
feat: improve diff sort

jsfez:


apps/database/src/models/ScreenshotDiff.ts
```diff
@@ -109,15 +109,17 @@ export class ScreenshotDiff extends Model {
...
+      THEN 30 -- removed
     WHEN "baseScreenshotId" IS NULL \
       AND "compareScreenshot"."name" ~ '${ScreenshotDiff.screenshotFailureRegexp}' \
       THEN 0 -- failure
     WHEN "baseScreenshotId" IS NULL  \
-      THEN 2 -- added
+      THEN 20 -- added
+    WHEN "score" IS NOT NULL AND "score" > 0 and "stabilityScore" < 60 \
+      THEN 11 -- changed and flaky
     WHEN "score" IS NOT NULL AND "score" > 0 \
```
gregberge, line 120:
```suggestion
    WHEN "score" IS NOT NULL  -- changed
        THEN (CASE
          WHEN "stabilityScore" < 60
            THEN 11 -- flaky
          ELSE 10 -- not flaky
        END) 
    WHEN "score" IS NOT NULL AND "score" > 0 \
```

gregberge: We can do the same optimization for failure / added. it avoid testing condition twice.

gregberge (changes requested):

apps/database/src/models/ScreenshotDiff.ts
```diff
@@ -107,17 +107,21 @@ export class ScreenshotDiff extends Model {
...
-      THEN 1 -- changed
-    ELSE 4 -- unchanged
+  static sortDiffByStatus = `CASE 
+    WHEN "compareScreenshotId" IS NULL 
+      THEN 30 -- removed 
+    WHEN "baseScreenshotId" IS NULL 
+      AND "compareScreenshot"."name" ~ '${ScreenshotDiff.screenshotFailureRegexp}' 
+      THEN 0 -- failure 
+    WHEN "baseScreenshotId" IS NULL 
+      THEN 20 -- added
```
gregberge:
```suggestion
    WHEN "baseScreenshotId" IS NULL
      THEN (CASE
        WHEN "compareScreenshot"."name" ~ '${ScreenshotDiff.screenshotFailureRegexp}'
          THEN 0 -- failure 
        ELSE 20 -- added 
      END)
```

gregberge: We can apply the same method here.

gregberge (approved):

PR #739, opened 2023-03-18
feat: add flaky list

jsfez:


apps/graphql/src/definitions/Test.ts
```diff
@@ -0,0 +1,132 @@
...
+      muted: Boolean!
+      muteUntil: String
+    ): MuteUpdateTest!
+  }
+`;
+
+export const resolvers = {
+  Test: {
+    lastSeen: async (test: Test) => {
+      const lastScreenshotDiff = await test.$relatedQuery("lastScreenshotDiff");
```
gregberge, line 68: Query n+1, use loader.

apps/graphql/src/definitions/Test.ts
```diff
@@ -0,0 +1,132 @@
...
+      return lastScreenshotDiff?.createdAt ?? null;
+    },
+    stabilityScore: async (test: Test) => {
+      const lastScreenshotDiff = await test.$relatedQuery("lastScreenshotDiff");
```
gregberge, line 72: Query n+1, use loader.

apps/app/src/pages/Repository/Tests.tsx
```diff
@@ -0,0 +1,563 @@
...
+
+const BuildNameField = ({ test }: { test: Test }) => {
+  return test.buildName !== "default" ? (
+    <div className="whitespace-nowrap">Build: {test.buildName}</div>
+  ) : null;
+};
+
+const LastSeenField = ({ test }: { test: Test }) => {
+  return (
+    <MagicTooltip tooltip={`Last seen: ${moment(test.lastSeen).format("LLL")}`}>
```
gregberge, line 139: `lastSeen` can be null, I fixed the generated types. It was considered as `any` that's why you didn't catch it.

Screenshot 2023-03-23 at 21 15 35

Ironical that the only wrong type results in a bug.

apps/app/src/pages/Repository/Tests.tsx
```diff
@@ -0,0 +1,563 @@
...
+        <input
+          type="checkbox"
+          checked={selectedTests.has(test.id)}
+          onChange={(e) => toggleTestSelection(test.id, e.target.checked)}
+        />
+      </div>
+      <div className="w-95 flex grow gap-4">
+        <Thumbnail screenshot={test.screenshot} />
+        <div className="flex flex-col justify-start gap-1">
+          <div className="flex min-h-[28px] items-start gap-2">
```
gregberge, line 246: We don't use px, but also why min-height 28?

gregberge (changes requested):
Screenshot 2023-03-23 at 21 19 02

- Graph --- 7 days is weird, why not just graph and a tooltip to precise it?



Screenshot 2023-03-23 at 21 20 30

- Time icon doesn't give any additional info, please remove it

Screenshot 2023-03-23 at 21 23 33

- The checkbox to check all is confusing since you use an infinite list. Does it check all the tests or just the displayed one? Anyway there is no need to bulk on all screenshots for any action. So please remove it.

Screenshot 2023-03-23 at 21 24 39

- If disabled, then the tooltip should be "To mute a test, first select one."

- The hover effect on the line is not good, if you want to add a hover effect it should have a click action. Please remove it while we don't have any "Test detail page" yet.

- The daily variation graph does not display anything if there is no test history, please show a message to tell the user about it.

Good job!

apps/app/src/pages/Repository/Tests.tsx
jsfez: I made an adjustment to ensure that the date remains in place even when a FlakyChip is displayed. To achieve this, I added a minimum height of 28px. Since Tailwind's utility classes don't include a min-h-7 option, I used pixels instead.

gregberge: Then use rem unit since Tailwind uses rem too.

PR #749, opened 2023-03-28
feat: add flaky dropdown

jsfez:
Capture d’écran 2023-03-28 à 17 23 15


gregberge (changes requested):
Only a link at this place. It is not good to take this kind of decision here.

PR #758, opened 2023-04-17
feat: add repository members

jsfez:


apps/app/src/pages/Repository/RepositorySettings.tsx
```diff
@@ -267,16 +290,68 @@ const VisibilityCard = ({ repository }: { repository: Repository }) => {
...
+  const usersByPage = 3;
 
   if (!ownerLogin || !repositoryName) {
     return <NotFound />;
   }
 
   if (!hasWritePermission) {
-    return <NotFound />;
+    return <Forbidden link={{ to: "..", label: "Back to repository" }} />;
   }
```
gregberge, line 355: We have to separate read & write if you do that.

apps/app/src/pages/Repository/index.tsx
```diff
@@ -19,21 +19,32 @@ const RepositoryQuery = graphql(`
     repository(ownerLogin: $ownerLogin, repositoryName: $repositoryName) {
       id
       permissions
+      tests(first: 1, after: 0) {
+        pageInfo {
+          totalCount
+        }
+      }
```
gregberge, line 26: It is not perf enough. You should add `hasTests` query. Counting all tests is not required to see if there is one test.

apps/app/src/ui/Pagination.tsx
```diff
@@ -0,0 +1,47 @@
...
+  handlePageChange: (after: number) => void;
+}
+
+export const Pagination: React.FC<PaginationProps> = ({
+  pageInfo,
+  first,
+  after,
+  handlePageChange,
+}) => {
+  const currentPage = after / first + 1;
```
gregberge, line 19: Not necessary to compute that if you render `null` next line.

gregberge (changes requested):

PR #764, opened 2023-04-24
fix tests 3

jsfez:
- test: add tests seeds
- fix(tests): fix tests layout
- feat: improve tests query


apps/app/src/containers/Layout.tsx
```diff
@@ -23,8 +23,8 @@ export const Main = forwardRef<HTMLElement, { children: React.ReactNode }>(
 export const Layout = (props: { children: React.ReactNode }) => {
   const fullSize = useMatch("/:accountSlug/:projectName");
   return (
-    <div className={clsx(fullSize && "h-screen", "min-h-screen")}>
-      <header>
+    <div className={clsx(fullSize && "h-screen", "flex min-h-full flex-col")}>
+      <header className="flex-shrink-0">
```
gregberge, line 27: Layout was broken that's why I made the change. I will fix it...

gregberge: If you revert it, then there is lot of other `shrink-0` to add.

gregberge (changes requested):

apps/graphql/src/definitions/Project.ts
```diff
@@ -33,6 +34,8 @@ export const typeDefs = gql`
     builds(first: Int!, after: Int!): BuildConnection!
     "A single build linked to the repository"
     build(number: Int!): Build
+    "Project has Tests"
+    hasTests: Boolean!
```
gregberge, line 38: I removed it because finally it is not the philosophy of GraphQL. If request is perf the totalCount should be perf too.

apps/app/src/containers/Layout.tsx
gregberge: I removed it on sub nav and tabs I think. With the "legacy" layout you are puting back, every elements in the page should have `shrink-0`. Else they will be "compacted" if there is a big list.

Anyway I removed it for a reason and puting it back is not the good solution...

gregberge (approved):

PR #774, opened 2023-04-27
feat: send consumption to stripe

jsfez:


apps/database/src/models/Build.ts
```diff
@@ -144,6 +145,16 @@ export class Build extends Model {
     return this.reload(queryContext);
   }
 
+  async $getAccount() {
```
gregberge, line 148: Related query does the same thing with throwIfNotFound().

gregberge (changes requested):

apps/database/src/models/Build.ts
```diff
@@ -144,6 +145,13 @@ export class Build extends Model {
...
```
gregberge, line 148: Retirer la méthode.

gregberge (changes requested):

apps/web/src/api/stripe.ts
```diff
@@ -164,6 +164,40 @@ const findSessionPlan = async (initialSession: Stripe.Checkout.Session) => {
...
+    }
+
+    const subscriptions = await stripe.subscriptions.list({
+      customer: stripeCustomerId,
+      expand: ["items"],
+      limit: 1,
+    });
+    const subscription = subscriptions.data[0];
+    const meteredPricing =
+      subscription?.items.data[0]?.price.recurring?.usage_type === "metered";
```
gregberge, line 181: If the user has several subscriptions?

apps/web/src/api/stripe.ts
```diff
@@ -164,6 +164,40 @@ const findSessionPlan = async (initialSession: Stripe.Checkout.Session) => {
...
+      subscription.id,
+      { action: "set", quantity }
+    );
+    return record.quantity;
+  } catch (e) {
+    throw new Error(
+      `error updating stripe usage: ${
+        e instanceof Error ? e.message : String(e)
+      }`
+    );
```
gregberge, line 197: Use `cause`.

apps/web/src/api/stripe.ts
```diff
@@ -164,6 +164,40 @@ const findSessionPlan = async (initialSession: Stripe.Checkout.Session) => {
...
+    if (!meteredPricing) {
+      return 0;
+    }
+
+    const quantity = await account.getScreenshotsCurrentConsumption();
+    const record = await stripe.subscriptionItems.createUsageRecord(
+      subscription.id,
+      { action: "set", quantity }
+    );
+    return record.quantity;
```
gregberge, line 191: You don't use that, why returning it then?

gregberge (changes requested):

apps/web/src/api/stripe.ts
jsfez: We don't provide multiple active subscription

gregberge: If prefer to request only `active` without `limit` if we found `> 1` then we throw.

apps/web/src/api/v2/builds/update.ts
```diff
@@ -128,6 +128,11 @@ const handleUpdateParallel = async ({
 
   if (complete) {
     await buildJob.push(build.id);
+    const account = await Project.relatedQuery<Account>("account")
```
gregberge, line 131: You are in a `transaction` and not passing `trx`. It should break tests.

gregberge (changes requested):

apps/web/src/api/v2/builds/update.ts
jsfez: Actually this line is out of transaction so I don't need trx. But do you think it should be included in a transaction? By the way, do you think there is a better place to call stripe update ?

gregberge: Yeah I think the build process before marking the build "complete" is a better place. What do you think?

apps/build/src/job.ts
```diff
@@ -15,6 +16,13 @@ export const performBuild = async (build: Build) => {
       .map(({ id }) => screenshotDiffJob.push(id))
   );
 
+  const account = await Project.relatedQuery<Account>("account")
+    .for(build.projectId)
+    .first()
```
gregberge, line 21: no need

apps/stripe/src/client.ts
```diff
@@ -0,0 +1,213 @@
...
+      customer: stripeCustomerId,
+      expand: ["data.items"],
+    });
+    if (subscriptions.data.length > 1) {
+      throw new Error(
+        `Can't update usage. Stripe return multiple active subscriptions for account ${account.id}`
+      );
+    }
+
+    const subscription = subscriptions.data[0];
```
gregberge, line 51: Add `if(!subscription) ...`

apps/stripe/src/client.ts
```diff
@@ -0,0 +1,213 @@
...
+    const meteredPricing =
+      subscription?.items.data[0]?.price.recurring?.usage_type === "metered";
+    if (!meteredPricing) {
+      return 0;
```
gregberge, line 55: `return null` instead of return 0 to distinguish the fact that you update to 0 or you didn't update.

apps/stripe/src/client.test.ts
```diff
@@ -22,6 +25,20 @@ const getOrCreateActivePurchase = async (account: Account, plan: Plan) => {
...
+    customer: customerId,
+  });
+  if (subscriptions.data.length > 0) {
+    return subscriptions.data[0] as Stripe.Subscription;
+  }
+  const subscription = await stripe.subscriptions.create({
+    customer: customerId,
+    items: [{ price: priceId }],
+  });
+  return subscription as Stripe.Subscription;
```
gregberge, line 39: why do we need this casting?

apps/stripe/src/client.test.ts
```diff
@@ -22,6 +25,20 @@ const getOrCreateActivePurchase = async (account: Account, plan: Plan) => {
   });
 };
 
+const getOrCreateSubscription = async (customerId: string, priceId: string) => {
+  const subscriptions = await stripe.subscriptions.list({
+    customer: customerId,
+  });
+  if (subscriptions.data.length > 0) {
+    return subscriptions.data[0] as Stripe.Subscription;
+  }
```
gregberge, line 34: The guard with the error when we have multiple should be there too.

apps/web/src/api/stripe.ts
```diff
@@ -1,313 +1,22 @@
...
-  }
-};
 
 router.post(
   "/stripe/event-handler",
   bodyParser.raw({ type: "application/json" }),
   async (req: express.Request, res: express.Response): Promise<void> => {
     let event: Stripe.Event;
-    const signature = req.headers["stripe-signature"] as string;
+    const signature: string = req.headers["stripe-signature"] as string;
```
gregberge, line 19:
```suggestion
    const signature = req.headers["stripe-signature"];
    if (!signature) {
      throw ...
    }
```

gregberge (changes requested):

apps/build/src/job.ts
jsfez: Heu... without `.first()`, it return an array of accounts. So, I think it's useful

apps/stripe/src/client.test.ts
```diff
@@ -36,6 +38,25 @@ const getOrCreateActivePurchase = async (account: Account, plan: Plan) => {
...
+  const subscriptions = await stripe.subscriptions.list({
+    customer: customerId,
+  });
+  if (subscriptions.data.length > 1) {
+    throw new Error(`Stripe return multiple active subscriptions`);
+  }
+
+  if (subscriptions.data.length > 0) {
+    return subscriptions.data[0] as Stripe.Subscription;
+  }
```
gregberge, line 51:
```suggestion
  const existingSubscription = subscriptions.data[0]
  if (existingSubscription) {
    return existingSubscription;
  }
```

gregberge (changes requested):
Last one to avoid casting.

PR #775, opened 2023-04-27
fix job isolation

gregberge:
- refactor: simplify builder
- chore(job): error logging isolation
- fix(sync): do not insert empty array

### DIFF
--- a/apps/database/src/models/GithubRepository.ts
+++ b/apps/database/src/models/GithubRepository.ts
@@ -62,7 +62,7 @@ export class GithubRepository extends Model {
           to: "github_installations.id",
         },
         modify(builder) {
-          return builder.findOne({ deleted: false });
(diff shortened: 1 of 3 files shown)


apps/database/src/models/GithubRepository.ts
```diff
@@ -62,7 +62,7 @@ export class GithubRepository extends Model {
           to: "github_installations.id",
         },
         modify(builder) {
-          return builder.findOne({ deleted: false });
+          builder.where("github_installations.deleted", false);
```
jsfez, line 65: It could return multiple activeInstallation. It not the expected result

gregberge: Nop, it is already a `hasOne` relation.

gregberge: I tested it.

PR #809, opened 2023-05-01
feat: store trial end date

jsfez:


apps/database/db/structure.sql
```diff
@@ -475,7 +475,7 @@ CREATE TABLE public.github_synchronizations (
     "createdAt" timestamp with time zone NOT NULL,
     "updatedAt" timestamp with time zone NOT NULL,
     "jobStatus" public.job_status NOT NULL,
-    "githubInstallationId" bigint
```
jsfez, line 478: @gregberge look like this field is not stable between your dumps and mine

gregberge: Look at the production and it will be the good version.

If I have a problem on my database I will fix it. If it is like production it is good then.

PR #810, opened 2023-05-01
feat: remove limit for usage based plans

jsfez:


apps/stripe/src/client.ts
```diff
@@ -45,19 +45,15 @@ export const updateStripeUsage = async ({
     });
     if (subscriptions.data.length > 1) {
       throw new Error(
-        `Can't update usage. Stripe return multiple active subscriptions for account ${account.id}`
+        `Stripe return multiple active subscriptions for account ${account.id}`
```
gregberge, line 48: returns

apps/stripe/src/client.ts
```diff
@@ -45,19 +45,15 @@ export const updateStripeUsage = async ({
...
     if (!subscription) {
-      return null;
-    }
-
-    const meteredPricing =
-      subscription.items.data[0]?.price.recurring?.usage_type === "metered";
-    if (!meteredPricing) {
-      return null;
+      throw new Error(
+        `Stripe return no active subscriptions for account ${account.id}`
```
gregberge, line 55: returns

apps/web/src/api/v2/util.ts
```diff
@@ -56,7 +56,8 @@ export const createBuild = async ({
   if (!isPublic) {
     const account = await req.authProject.$relatedQuery("account", trx);
     const hasExceedLimit = await account.hasExceedScreenshotsMonthlyLimit();
-    if (hasExceedLimit) {
+    const hasUsageBasedPlan = await account.hasUsageBasedPlan();
```
gregberge, line 59: I think methods are too abstract. Then you fetch the same thing multiple times. Not a big deal but maybe to refactor later.

gregberge (changes requested):

apps/stripe/src/client.ts
jsfez, line 55:
```suggestion
        `Stripe returns no active subscriptions for account ${account.id}`
```

apps/stripe/src/client.ts
jsfez, line 48:
```suggestion
        `Stripe returns multiple active subscriptions for account ${account.id}`
```

PR #828, opened 2023-05-07
feat: update team settings

### LINKED ISSUE #826: Team settings rework
Dependent of issue [#822](https://github.com/argos-ci/argos/issues/822)

- [x] Rename page name to « team settings »
- [x] Button « Early trial end »
- [x] Remove checkout link

Plan statuses :
  - [x] No purchase (legacy)
  - [x] Trial
    - [x] Missing payment method
    - [x] Trial canceled
    - [x] Subscription will start after trial

  - [x] Trial expired
  - [x] Pro plan
  - [x] Pro plan canceled

jsfez:
Fix #826


apps/graphql/src/definitions/Purchase.ts
```diff
@@ -23,5 +38,50 @@ export const resolvers: IResolvers = {
     account: async (purchase, _args, ctx) => {
       return ctx.loaders.Account.load(purchase.accountId);
     },
+    plan: async (purchase) => {
+      return purchase.$relatedQuery("plan");
```
gregberge, line 42: Loader better.

apps/graphql/src/definitions/Team.ts
```diff
@@ -151,6 +154,12 @@ export const resolvers: IResolvers = {
       return team.$getInviteLink();
     },
   },
+  periodStartDate: async (account) => {
+    return account.$getCurrentConsumptionStartDate();
+  },
+  periodEndDate: async (account) => {
+    return account.$getCurrentConsumptionEndDate();
```
gregberge, line 161: Perf problems, multi fetching etc..

gregberge: Also it should be on `Account` type, not on `Team` and `User`.

apps/app/src/ui/StripePortalLink.tsx
```diff
@@ -1,39 +1,43 @@
...
-
-import config from "@/config";
+import { clsx } from "clsx";
 
 import { Anchor, anchorClassNames } from "./Link";
 
+export const ContactLink = () => {
+  return (
+    <>
+      Contact Argos support via{" "}
```
gregberge, line 9: Is "via" correct in english? I think "through" is better.

apps/app/src/ui/StripePortalLink.tsx
```diff
@@ -1,39 +1,43 @@
...
+      <Anchor href="https://discord.gg/WjzGrQGS4A" external>
+        Discord
+      </Anchor>{" "}
+      or{" "}
+      <Anchor href="mailto:contact@argos-ci.com" external>
+        by e-mail
```
gregberge, line 15:
```suggestion
        by email
```

apps/app/src/ui/Chip.tsx
```diff
@@ -14,7 +14,7 @@ export interface ChipProps
   extends Omit<React.ComponentProps<"div">, "className"> {
   color?: ChipColor;
   icon?: React.ComponentType<any> | null;
-  scale?: "sm" | "md" | undefined;
+  scale?: "xs" | "sm" | "md" | undefined;
```
gregberge, line 17: Implementation?

gregberge (changes requested):

apps/app/src/ui/StripePortalLink.tsx
jsfez: While "via" is technically correct in English, if it bothers you, I can use "on" instead.

apps/app/src/ui/Chip.tsx
jsfez: removed

apps/graphql/src/definitions/Team.ts
gregberge: I will work on it, let it like that.

PR #842, opened 2023-05-17
fix: has usage-based plan

jsfez:


apps/database/src/models/Account.ts
```diff
@@ -237,15 +237,15 @@ export class Account extends Model {
...
     if (activePurchase.$isTrialActive()) {
       return false;
     }
 
-    return activePurchase.plan.usageBased;
+    const plan =
+      activePurchase.plan ??
+      (await activePurchase.$relatedQuery<Plan>("plan").first());
+
+    return Boolean(plan?.usageBased);
```
gregberge, line 248: Why ?.

apps/database/src/models/Account.ts
```diff
@@ -237,15 +237,15 @@ export class Account extends Model {
...
-      throw new Error("Invariant plan not loaded");
-    }
-
     if (activePurchase.$isTrialActive()) {
       return false;
     }
 
-    return activePurchase.plan.usageBased;
+    const plan =
+      activePurchase.plan ??
```
gregberge, line 245: Incoherent if you use .? after, it should be === undefined if plan could be null.

gregberge (changes requested):

PR #855, opened 2023-05-22
fix purchase actions

jsfez:
This PR has several objectives : 

1. Use the new `account.purchaseStatus` to remove logic from the frontend. It allow us to handle every purchase statuses in `PaymentBanner` and `PlanCard` components.

2. Always display "manage subscription" button when a purchase has been performed on an account.

3. Re-use the `stripeCustomerId` for new purchases when it has been set on an account.

Commits: 
- fix: plan card actions
- feat: add customer to stripe session checkout
- feat: udpate payment banner


jsfez:
This PR has several objectives : 

1. Use the new `account.purchaseStatus` to remove logic from the frontend. It allow us to handle every purchase statuses in `PaymentBanner` and `PlanCard` components.

2. Always display "manage subscription" button when a purchase has been performed on an account.

3. Re-use the `stripeCustomerId` for new purchases when it has been set on an account.

gregberge (approved):

PR #862, opened 2023-05-23
fix(builds): fix layout

jsfez:


gregberge (approved):
Without `min-h-0` it should not work well but I trust you.

gregberge (approved):
Without `min-h-0` it should not work well but I trust you.

PR #869, opened 2023-05-23
feat: swap between screenshots

### LINKED ISSUE #865: Add "swap to baseline <=> change" button

jsfez:
fix #865


gregberge (changes requested):
Nice but it should be possible through the toolbar.

Your model is not correct, we only want : 

- view: "split" | "baseline" | "changes"

jsfez:
I have to think about a way to display this button then

PR #888, opened 2023-05-30
feat: crawl strategy

### LINKED ISSUE #881: Create crawler strategy
The crawler should retrieve the 30 first link that are accessible from homepage.
The crawler should use [cheerio](https://github.com/cheeriojs/cheerio) 

[Crawler code](https://github.com/argos-ci/argos/blob/main/apps/crawl/src/job.ts)

jsfez:
fix #881

### DIFF
--- a/apps/crawl/src/job.ts
+++ b/apps/crawl/src/job.ts
@@ -1,15 +1,92 @@
+import axios from "axios";
+import * as cheerio from "cheerio";
+
 import { job as captureJob } from "@argos-ci/capture";
 import { Capture, Crawl } from "@argos-ci/database/models";
 import { createModelJob } from "@argos-ci/job-core";
 
-export const performCrawl = async (crawl: Crawl) => {
-  const capture = await Capture.query().insertAndFetch({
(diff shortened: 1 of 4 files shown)


apps/crawl/src/job.ts
```diff
@@ -1,15 +1,92 @@
...
+    .insert(
+      urls.map((url) => ({
+        jobStatus: "pending" as const,
+        crawlId: crawl.id,
+        url,
+      }))
+    );
+  captures.forEach((capture) => {
+    captureJob.push(capture.id);
+  });
```
jsfez, line 89: @gregberge Could you please confirm if this is the expected behavior? I'm unsure about the functionality of the jobs and would appreciate your insights.

gregberge: Yeah

gregberge (changes requested):
Your algorithm is recursive, we want only the 30 first links on homepage. Code looks very complex to just list links on homepage, filter absolutes one.

PR #921, opened 2023-07-02
feat(subscription): start trial server-side

### LINKED ISSUE #915: Start Trial server-side at Team creation
When a Team is created, the trial should have started server-side.

gregberge:
Closes #915

- Automatic trial creation (if the user has an email)
- No extra step from subscription

Next step, change the GitHub permission scope to get the email of all our users.


apps/app/src/containers/PlanCard.tsx
```diff
@@ -572,12 +565,14 @@ export const PlanCard = (props: { account: AccountFragment }) => {
           <ContactLink />
         ) : (
           <div className="flex items-center justify-between">
-            <div>
-              <ManageSubscriptionButton
-                stripeCustomerId={stripeCustomerId ?? ""}
-                paymentProvider={paymentProvider ?? null}
-              />
-            </div>
+            {stripeCustomerId && (
```
jsfez, line 568: Github subscriber won't be able to manage their subscriptions with this code

jsfez, line 568: Github subscribers won't be able to manage their subscriptions with this code

jsfez (changes requested):

PR #942, opened 2023-07-14
feat: various ui improvements

jsfez:


apps/app/src/ui/Breadcrumb.tsx
```diff
@@ -51,9 +51,9 @@ export const BreadcrumbMenuButton = (props: { state: MenuState }) => {
   return (
     <MenuButton
       state={props.state}
-      className="cursor-default p-1 text-on-light transition hover:text-on aria-expanded:text-on"
+      className="cursor-pointer rounded-md border border-border p-0.5 text-on-light transition hover:text-on aria-expanded:bg-slate-800 aria-expanded:text-on"
```
gregberge, line 54:
```suggestion
      className="cursor-default rounded-md border border-border p-0.5 text-on-light transition hover:text-on aria-expanded:bg-slate-800 aria-expanded:text-on"
```

gregberge: No pointer if the user stays on the same page.

gregberge (changes requested):

PR #957, opened 2023-07-18
fix(expired): fix expired status message

gregberge:


jsfez (approved):

jsfez:
Did you break our cooking status?

PR #1002, opened 2023-09-02
feat(settings): update settings tabs labels

jsfez:


gregberge (changes requested):
I don't get the point. Why not Personal Projects?

The space is defined by the Breadcrumb header.

jsfez:
Argos has three "Settings" tabs in the same position, leading to different pages: Team, Personal, and Project Settings. Even as a co-creator, I find this confusing. To improve UX, I suggest renaming these tabs for clarity.

jsfez:
Do you validate the PR or do you think we should address this point by an other way?

PR #1028, opened 2023-09-16
feat(build): improve active screenshot visibility

jsfez:


gregberge (changes requested):
I tested it, I am not convinced. We can use a biggest border to improve this. I can work on it if it is a real issue, but the bg does not really improve it. I barely see the difference on my Macbook.

jsfez:
If you can barely see the background on a MB, the change will be invisible on a screen that is not as good quality. So my fix may not be the best one

jsfez:
> I tested it, I am not convinced. We can use a biggest border to improve this. I can work on it if it is a real issue, but the bg does not really improve it. I barely see the difference on my Macbook.

If you can barely see the background on a MB, the change will be invisible on a screen that is not as good quality. So my fix may not be the best one

gregberge (approved):

PR #1040, opened 2023-09-23
feat(s3): use diff fingerprint as s3 key

jsfez:


apps/screenshot-diff/src/computeScreenshotDiff.ts
```diff
@@ -123,21 +124,26 @@ export const computeScreenshotDiff = async (
...
     if (diffResult.score > 0) {
+      const diffKey = await hashFile(diffResult.filepath);
+      const duplicateFile = await File.query().findOne({ key: diffKey });
       const diffImage = new S3ImageFile({
         s3,
         bucket: bucket,
         filepath: diffResult.filepath,
+        ...(duplicateFile && { key: duplicateFile.key }),
       });
       const key = await diffImage.upload();
```
gregberge, line 135: I think you will upload it anyway. Not good I think.

gregberge (changes requested):

apps/screenshot-diff/src/computeScreenshotDiff.ts
jsfez: Yes and no. I won't be re-uploaded but it will throw an error. You are right, it's not good enough

Capture d’écran 2023-09-23 à 19 19 59

apps/screenshot-diff/src/computeScreenshotDiff.ts
```diff
@@ -51,6 +66,135 @@ async function completeFile({
...
+      });
+      return;
+    }
+
+    await diffImage.upload();
+    await transaction(async (trx) => {
+      const diffFile = await File.query(trx)
+        .insert({ key, width: diffResult.width, height: diffResult.height })
+        .returning("*");
+      await unlinkFileAndPatchScreenshotDiff(
```
gregberge, line 124: No need to do that in the lock, only the file creation. Updating the diff should not be done in the lock. We don't care about the transaction here.

gregberge (changes requested):

gregberge (approved):

gregberge (approved):

PR #1041, opened 2023-09-25
feat(diff-list): sort diffs by s3Id

jsfez:


gregberge (changes requested):
If you do that, the order will be totally random, not based on named any more.

gregberge:
@jsfez I close it.

PR #1052, opened 2023-09-30
feat: add diffs group

jsfez:


gregberge (changes requested):
You can miss one at this step, you should do it at the end of the diff.

apps/screenshot-diff/src/computeScreenshotDiff.ts
```diff
@@ -67,20 +67,41 @@ async function ensureFileDimensionsAndPreload({
...
+  buildId: string;
+}) {
+  const file = await File.query().findOne({ key });
+  if (!file) return null;
+
+  await ScreenshotDiff.query()
+    .where({ buildId, s3Id: key, group: null })
+    .patch({ group: key });
+  return file;
+}
```
gregberge, line 87: This is not needed.

apps/screenshot-diff/src/computeScreenshotDiff.ts
```diff
@@ -229,14 +260,28 @@ export const computeScreenshotDiff = async (
...
+    .where({ buildId })
+    .whereNotNull("group");
+
+  await Promise.all(
+    diffsGroups.map(async (diffsGroup) =>
+      ScreenshotDiff.query()
+        .where({ buildId, s3Id: diffsGroup.s3Id, group: null })
+        .patch({ group: diffsGroup.group }),
+    ),
+  );
```
gregberge, line 275:
```ts
ScreenshotDiff.query()
        .where({ buildId, s3Id: screenshotDiff.s3Id, group: null })
        .patch({ group: diffsGroup.group })
```

gregberge (changes requested):
Remove all code except the patch at the end of the compute, it is enough.

apps/screenshot-diff/src/computeScreenshotDiff.ts
gregberge, line 275: We don't want to patch all groups, just the group concerning our screenshot diff.

```ts
const similarDiffCount = await ScreenshotDiff.query().where({ buildId, s3Id: screenshotDiff.s3Id }).resultSize()
if (similarDiffCount > 1) {
  await ScreenshotDiff.query().where({ buildId, s3Id: screenshotDiff.s3Id, group: null }).patch({ group: screenshotDiff.s3Id });
}
```

gregberge (changes requested):

jsfez:
I agree. You version is easier to maintain

apps/database/db/structure.sql
```diff
@@ -401,7 +401,7 @@ ALTER SEQUENCE public.github_installations_id_seq OWNED BY public.github_install
 --
 
 CREATE TABLE public.github_pull_requests (
-    id bigint NOT NULL,
+    id integer NOT NULL,
```
gregberge, line 404: Not good

gregberge (changes requested):
Something weird in the structure but the rest is OK 👌. Maybe just check the type in production, report it in your db and in the structure. But I think it should be bigint.

PR #1053, opened 2023-09-30
feat(build): group diffs

jsfez:
Capture d’écran 2023-10-01 à 11 45 01
Capture d’écran 2023-10-01 à 11 45 08
Capture d’écran 2023-10-01 à 11 45 14


gregberge (commented):
@jsfez can I check the result locally with seeds?

gregberge (approved):

PR #1102, opened 2023-11-20
feat(gitlab): list subgroup projects

jsfez:
GitLab API doc: https://docs.gitlab.com/ee/api/groups.html#list-a-groups-projects


gregberge (changes requested):
Not what is expected. We need to do things differently for GitLab.

PR #1120, opened 2023-12-03
chore: setup knip

gregberge:
Remove unused things, run on CI

https://knip.dev/


jsfez (approved):
Very cool tool !

knip.json
```diff
@@ -0,0 +1,40 @@
...
+  },
+  "webpack": {
+    "config": ["webpack.config.mjs"]
+  },
+  "vitest": {
+    "config": ["vitest.config.ts", "vitest.*.config.ts"],
+    "entry": ["**/*.e2e.test.ts?(x)", "**/*.test.ts?(x)"]
+  },
+  "exclude": ["classMembers"],
+  "ignore": ["prettier.config.js", "playwright.config.mjs", "examples"],
```
webpro, line 34: Two potential solutions for the problematic `prettier.config.js`. You could change the `require()` calls to `require.resolve()` or convert it to ESM, so you would not need to ignore it here, and thus also not the items in `ignoreDependencies` below.

knip.json
```diff
@@ -0,0 +1,40 @@
...
+      "entry": ["src/index.tsx"],
+      "ignore": ["src/gql"],
+      "ignoreDependencies": ["webpack-cli"]
+    },
+    "apps/email": {
+      "entry": ["emails/*.tsx"]
+    }
+  },
+  "webpack": {
+    "config": ["webpack.config.mjs"]
```
webpro, line 27: Knip v3.5.0 has the `.mjs` included so this could be removed after updating.

knip.json
```diff
@@ -0,0 +1,40 @@
...
+  },
+  "vitest": {
+    "config": ["vitest.config.ts", "vitest.*.config.ts"],
```
webpro, line 30: Seeing this pattern more often, Knip v3.5.0 uses `vitest*.config.ts` so this could be removed if you update.

knip.json
```diff
@@ -0,0 +1,40 @@
...
+    "entry": ["**/*.e2e.test.ts?(x)", "**/*.test.ts?(x)"]
```
webpro, line 31: You don't need to add this manually, Knip also checks Vitest config.

knip.json
```diff
@@ -0,0 +1,40 @@
...
+      "entry": [
+        "src/processes/proc/*.ts",
+        "src/build/bin/*.ts",
+        "src/database/testing/index.ts",
+        "src/graphql/definitions/*.ts"
+      ],
+      "ignore": ["db", "knexfile.js", "src/graphql/__generated__"],
+      "ignoreDependencies": ["pg"],
+      "paths": {
+        "@/*": ["./src/*"]
```
webpro, line 14: When I remove this I get the same results. I think this `paths` config can be removed, because `tsconfig.json` has it (and Knip uses that as well).

knip.json
```diff
@@ -0,0 +1,40 @@
...
+      }
+    },
+    "apps/frontend": {
+      "entry": ["src/index.tsx"],
```
webpro, line 18: This is a default setting that could be removed.

knip.json
```diff
@@ -0,0 +1,40 @@
...
+      "ignore": ["src/gql"],
+      "ignoreDependencies": ["webpack-cli"]
```
webpro, line 20: This should have been detected properly, fixed the bug in Knip v3.5.0 so you don't need to ignore `webpack-cli` if you decide to update.

webpro (commented):
Took the liberty of an unsolicited review round, sorry if it's invasive. Sometimes I check a project to see what Knip can do to minimize config. Also, I fixed a few things in Knip v3.5.0 you can take advantage of.

gregberge:
@webpro thanks a lot for this review, I simplified the configuration with your suggestions!

webpro:
> @webpro thanks a lot for this review, I simplified the configuration with your suggestions!

Awesome, invaluable to see how it's used in projects. Every bit helps to make Knip better. Good luck with Argos!

PR #1123, opened 2023-12-05
feat: improve bad token error message

jsfez:


apps/backend/src/web/middlewares/repoAuth.ts
```diff
@@ -26,7 +26,7 @@ export const repoAuth = [
     if (!bearerToken) {
       throw new HttpError(
         401,
-        `Missing bearer token. Please provide a token in the Authorization header.`,
+        `Missing bearer token. Please provide a token in the Authorization header. See https://app.argos-ci.com`,
```
gregberge, line 29: I don't understand why you redirect to the app, it does not help.

gregberge (approved):

apps/backend/src/web/middlewares/repoAuth.ts
jsfez: I provide a direct link to add the project in Argos. What else would you want to display?

gregberge: I think it is not relevant here for this error but anyway this one never happens so.

PR #1136, opened 2023-12-28
fix: tokenless naming conflict

### LINKED ISSUE #1134: Bug: naming conflict in tokenless strategy
## Context
The tokenless strategy doesn't work if an other project has the same name as my current project.

## How to reproduce
1. Connect a GitHub repository to Argos.
2. Delete the project from Argos and GitHub.
3. Create a new GitHub repository with the same name.
4. Connect the new project to Argos.

**Issue:** New builds will fail because Argos try to connect with the old project.

## Solution
Improve the tokenless strategy.

jsfez:
fix #1134

### DIFF
--- a/apps/backend/src/web/middlewares/tokenless-strategies/github-actions.ts
+++ b/apps/backend/src/web/middlewares/tokenless-strategies/github-actions.ts
@@ -51,6 +51,7 @@ const strategy = {
       .withGraphJoined("[activeInstallation, projects]")
       .where("githubAccount.login", authData.owner)
       .findOne("github_repositories.name", authData.repository)
+      .orderBy("github_repositories.updatedAt", "desc")
       .first();
 
     if (


apps/backend/src/web/middlewares/tokenless-strategies/github-actions.ts
```diff
@@ -51,6 +51,7 @@ const strategy = {
       .withGraphJoined("[activeInstallation, projects]")
       .where("githubAccount.login", authData.owner)
       .findOne("github_repositories.name", authData.repository)
+      .orderBy("github_repositories.githubId", "desc")
```
gregberge, line 54: Not sure it works if I rename an old repository.

gregberge (changes requested):

apps/backend/src/web/middlewares/tokenless-strategies/github-actions.ts
jsfez: You can't rename an old repo. It has been deleted on GitHub

PR #1138, opened 2024-01-08
feat: allow to configure summary checks

gregberge:


apps/backend/db/structure.sql
```diff
@@ -403,7 +403,7 @@ ALTER SEQUENCE public.github_installations_id_seq OWNED BY public.github_install
 --
 
 CREATE TABLE public.github_pull_requests (
-    id bigint NOT NULL,
+    id integer NOT NULL,
```
jsfez, line 406: Why ?

jsfez (changes requested):

PR #1147, opened 2024-01-14
feat: paginate github repo list

jsfez:
CleanShot 2024-01-14 at 17 56 58@2x

CleanShot 2024-01-14 at 17 57 13@2x

CleanShot 2024-01-14 at 17 57 18@2x

CleanShot 2024-01-14 at 17 57 25@2x


gregberge (approved):
I think previous / next would have been enough but fine. Also could you please put 100 repos per page? I think it's better to avoid pagination if possible.

PR #1159, opened 2024-01-24
feat: support reference branch token

gregberge:
Related to https://github.com/argos-ci/argos-javascript/pull/105

### DIFF
--- a/apps/backend/src/web/api/v2/util.ts
+++ b/apps/backend/src/web/api/v2/util.ts
@@ -152,11 +152,34 @@ const createBuild = async (params: {
   return build;
 };
 
+/**
+ * Special token used to reference the reference branch of the project.
+ * Interpreted by Argos SDKs especially for GitHub Actions with "deployment_status" event.
+ */
+const REFERENCE_BRANCH = "__argos/reference-branch";
+
+/**
+ * Resolves the reference branch if a special token is used.
+ */
+async function resolveReferenceBranch(
+  referenceBranch: string,
+  project: Project,
+) {
+  if (referenceBranch === REFERENCE_BRANCH) {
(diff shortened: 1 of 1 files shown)


apps/backend/src/web/api/v2/util.ts
```diff
@@ -152,11 +152,34 @@ const createBuild = async (params: {
...
+ */
+async function resolveReferenceBranch(
+  referenceBranch: string,
+  project: Project,
+) {
+  if (referenceBranch === REFERENCE_BRANCH) {
+    return project.$getReferenceBranch();
+  }
+
+  return null;
```
jsfez, line 172: The previous value was `req.body.referenceBranch ?? null`.

The new value is `referenceBranch === REFERENCE_BRANCH ? project.$getReferenceBranch() : null`

It should be : `referenceBranch === REFERENCE_BRANCH ? project.$getReferenceBranch() : referenceBranch`

apps/backend/src/web/api/v2/util.ts
```diff
@@ -169,6 +192,6 @@ export const createBuildFromRequest = async ({
     prNumber: req.body.prNumber ?? null,
     prHeadCommit: req.body.prHeadCommit ?? null,
     referenceCommit: req.body.referenceCommit ?? null,
-    referenceBranch: req.body.referenceBranch ?? null,
+    referenceBranch: referenceBranch,
```
jsfez, line 195:
```suggestion
    referenceBranch: referenceBranch ?? null,
```

apps/backend/src/web/api/v2/util.ts
jsfez, line 172:
```suggestion
  return referenceBranch;
```

PR #1161, opened 2024-01-25
fix links in welcome email template

Nav-2d:
The email that is sent to new users contain incorrect links.

Tried to reference the correct pages based on the wording. Let me know if it needs polishing.


jsfez:
Hello @Nav-2d, thank you for your PR. It let me knows that the welcome e-mail was broken. Finally, I have changed more than the links: #1161

PR #1165, opened 2024-01-25
fix: diff layer position on firefox

jsfez:


apps/frontend/src/pages/Build/BuildDetail.tsx
```diff
@@ -216,15 +216,22 @@ const BaseScreenshot = ({ diff }: { diff: Diff }) => {
...
               width: diff.width,
               height: diff.height,
             })}
           />
           <img
-            className="absolute left-0 top-0"
+            className={clsx(
+              "absolute top-0",
+              contained && "max-h-full",
+              "left-0 right-0 mx-auto",
```
gregberge, line 233: Merge this with the first line

apps/frontend/src/pages/Build/BuildDetail.tsx
```diff
@@ -317,13 +324,19 @@ const CompareScreenshot = ({ diff }: { diff: Diff }) => {
           }
         >
           <img
-            className={clsx("absolute", visible && "opacity-disabled")}
+            className={clsx(
+              "absolute",
+              visible && "opacity-disabled",
+              // Firefox fixes
+              contained && "max-h-full",
+              "left-0 right-0 mx-auto",
```
gregberge, line 332: I don't think it works if the comparison screenshot is smaller than the diff overlay, but maybe we have the bug today.

gregberge (changes requested):
Are you sure that we need mx-auto everywhere like that? Looks weird.

gregberge:
This fix is valid but does not solve the underlying issue that the parent div is not correctly sized.

See #1167 for a better fix.

PR #1182, opened 2024-02-04
Read included screenshots count from Stripe

gregberge:
Read included screenshots directly from Stripe subscription to have a custom included screenshots number by Stripe price.


apps/frontend/src/pages/Build/OvercapacityBanner.tsx
```diff
@@ -25,10 +26,7 @@ export const OvercapacityBanner = memo(
     const { accountSlug } = props;
     const account = useFragment(AccountFragment, props.account);
     const { plan, consumptionRatio } = account;
-    const visible =
-      plan && typeof consumptionRatio === "number" && consumptionRatio >= 0.9;
-
-    if (!visible) {
+    if (!plan?.usageBased) {
```
jsfez, line 29: The overcapacity banner should not be displayed if consumptionRatio < 0.9

apps/backend/src/stripe/index.ts
```diff
@@ -148,30 +148,58 @@ const checkSubscriptionPaymentMethodFilled = async (
...
+    return null;
+  }
+  const { tiers } = await stripe.prices.retrieve(price.id, {
+    expand: ["tiers"],
+  });
+  invariant(tiers, "no tiers found in a tiered price");
+
+  // Take flat tiers only
+  const flatTiers = tiers.filter((tier) => tier.up_to && tier.flat_amount);
+  // Take the last one (the highest number of screenshots allowed)
```
jsfez, line 166: Are the tiers sorted by size ? The comment says "the highest" but you picked the last one. You should use a `Math.max()` instead of last index.

jsfez (changes requested):

jsfez (approved):

PR #1194, opened 2024-02-15
feat: support reference build triggered from PR

gregberge:
Also improve perf by minimizing the transaction process.


apps/backend/src/build/createBuildDiffs.ts
```diff
@@ -23,25 +23,19 @@ const getBuildType = ({
...
 
-  const baseScreenshotBucket = await getBaseScreenshotBucket({
-    build,
-    trx,
-  });
+  const baseScreenshotBucket = await getBaseScreenshotBucket(build);
 
   if (baseScreenshotBucket) {
-    await Build.query(trx)
+    await Build.query()
```
jsfez, line 34: If you remove the transaction here, the build can be updated several times in case or sharding, but it's ok

jsfez (approved):

PR #1195, opened 2024-02-16
fix(lock): improve redis lock algo

gregberge:
- Prevent memory leak by breaking Promise chaining
- Add random delay between retries to avoid clashing
- Set a random value and check it to avoid to be removed by another
  client
- Crash if process is longer than timeout

### DIFF
--- a/apps/backend/src/util/redis/lock.ts
+++ b/apps/backend/src/util/redis/lock.ts
@@ -1,9 +1,8 @@
-import { setTimeout } from "node:timers/promises";
 import type { createClient } from "redis";
 
 type RedisClient = ReturnType<typeof createClient>;
 
-const acquireLock = async ({
+const acquireLock = ({
   client,
   name,
   timeout,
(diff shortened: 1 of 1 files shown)


apps/backend/src/util/redis/lock.ts
```diff
@@ -14,45 +13,60 @@ const acquireLock = async ({
   timeout: number;
   retryDelay: number;
 }) => {
-  const result = await client.set(name, "1", {
-    PX: timeout,
-    NX: true,
-  });
+  return new Promise<string>((resolve, reject) => {
+    function tryAcquire() {
+      const rdn = Math.random().toString(36);
```
jsfez, line 18: Why did you use a base 36 here? To store a bigger number?

apps/backend/src/util/redis/lock.ts
```diff
@@ -14,45 +13,60 @@ const acquireLock = async ({
...
+      client
+        .set(name, rdn, {
+          PX: timeout,
+          NX: true,
+        })
+        .then((result) => {
+          if (result === "OK") {
+            resolve(rdn);
+          } else {
+            const adjustedTimeout = retryDelay + Math.ceil(Math.random() * 10);
```
jsfez, line 28: `Math.ceil(Math.random() * 10)` equal 0ms or 1ms. Is this what you expected?

PR #1196, opened 2024-02-18
fix: build list scroll

jsfez:


gregberge (changes requested):
Possible to avoid ! ?

jsfez:
I have update the code to remove the bang. I'll have to check on members' lists if there are regressions (I can't test it locally).

gregberge (approved):

PR #1197, opened 2024-02-19
chore: use bigint for "subscriptions" table columns

gregberge:


apps/backend/db/structure.sql
```diff
@@ -971,8 +971,8 @@ CREATE TABLE public.subscriptions (
     "planId" integer NOT NULL,
     provider text NOT NULL,
     "stripeSubscriptionId" character varying(255),
-    "accountId" integer NOT NULL,
-    "subscriberId" integer,
+    "accountId" bigint,
```
jsfez, line 974: Missing "not null"

jsfez (changes requested):

jsfez (approved):

PR #1245, opened 2024-03-26
chore: bump to node@20 volta

ryota-murakami:
I faced bellow message while setup project following CONTRIBUTING.md, so upgrade pinned volta's Node.js version as name as `engines` filed on `package.json`.


package.json
```diff
@@ -57,6 +57,6 @@
     "vitest": "^1.4.0"
   },
   "volta": {
-    "node": "18.17.1"
+    "node": "20.12.0"
```
gregberge, line 60: Better to put `"20"` as we want to use the latest of "20", not any version in particular.

gregberge, line 60: Better to put "20" as we want to use the latest of "20", not any version in particular.

gregberge (changes requested):

package.json
ryota-murakami: @gregberge Good idea but Volta doesn't allow it. 😭

```console
~/f/argos (fix/volta-node-version|✚1) [126]$ node -v
Volta error: Could not parse version "20"
```

PR #1258, opened 2024-04-07
chore: update scripts

gregberge:


jsfez (commented):
The files are called « queue pending job » but they request « error » only. Is there a mistake ?

gregberge:
@jsfez yeah but not really important, I have to refactor it later. I adapt them for each new case, so it's just to have base.

PR #1270, opened 2024-04-28
vite

gregberge:


apps/frontend/package.json
```diff
@@ -28,12 +28,11 @@
...
     "@types/react-dom": "^18.2.23",
     "@types/react-helmet": "^6.1.11",
+    "@vitejs/plugin-react": "^4.2.1",
     "ariakit": "^2.0.0-next.44",
-    "assets-webpack-plugin": "^7.1.1",
     "autoprefixer": "^10.4.19",
     "axios": "^1.6.8",
     "clsx": "^2.1.0",
-    "core-js": "^3.36.1",
     "css-loader": "^6.10.0",
```
wojtekmaj, line 36: Still needed?

apps/frontend/vite.config.ts
```diff
@@ -0,0 +1,64 @@
...
+      },
+    },
+    resolve: {
+      alias: [
+        {
+          find: "@",
+          replacement: fileURLToPath(new URL("./src", import.meta.url)),
+        },
+      ],
+    },
```
wojtekmaj, line 31: Should not be needed with package.json imports 🤔

apps/frontend/vite.config.ts
```diff
@@ -0,0 +1,64 @@
...
+    define: {
+      "process.env.NODE_ENV": JSON.stringify(mode),
+    },
```
wojtekmaj, line 34: Have a read: https://vitejs.dev/guide/env-and-mode 
Might be an interesting alternative to Node.js-mocking-era `process.env.*`.

apps/frontend/package.json
gregberge: Removed in another release thanks to [Knip](https://knip.dev/).

apps/frontend/vite.config.ts
gregberge: I will try it!

apps/frontend/vite.config.ts
gregberge: Lot of libraries still relies on process.env.NODE_ENV, I think it's better to stick on it for now. In my code not necessarily but for dependencies.

PR #1274, opened 2024-04-29
feat: open added & removed section by default

gregberge:


jsfez (approved):

jsfez:
I'm not sure about the removed, but OK

PR #1299, opened 2024-05-19
feat(account): delete account in a transaction

gregberge:
Handle account with a Slack installation

### DIFF
--- a/apps/backend/src/graphql/services/account.ts
+++ b/apps/backend/src/graphql/services/account.ts
@@ -1,5 +1,6 @@
 import { assertNever } from "@argos/util/assertNever";
 import { invariant } from "@argos/util/invariant";
+import { RelationExpression } from "objection";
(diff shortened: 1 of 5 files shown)


apps/backend/src/graphql/services/account.ts
```diff
@@ -8,62 +9,104 @@ import {
...
+ */
+export async function deleteAccount(args: {
   id: string;
   user: User | undefined | null;
-}) => {
+}): Promise<void> {
   const account = await getAdminAccount({
     id: args.id,
     user: args.user,
+    withGraphFetched: "[projects,slackInstallation]",
```
jsfez, line 47: It would be better to provide a list instead of a string here

gregberge: I prefer the string syntax notation

PR #1302, opened 2024-05-22
feat(ui): migrate to react-aria-components

gregberge:
Replace Radix & Ariakit by React Aria.


frontsideair:
Coming from the blog post, congrats on landing this PR. It's also nice that the code size is reduced after the migration, if you discount the lockfile controbution.

PR #1328, opened 2024-06-16
feat: support GitHub Actions partial retries

gregberge:
## Strategy

- We collect informations on the build: `runId` and `runAttempt`.
- We determine if it's a partial build or not. To determine that, we look at the jobs of the run, if two are similar in terms of date, then it's a partial retry.
- If it's a partial build, then we will finish this build when the run will be completed (we listen the GitHub webhook).

### DIFF
--- a/apps/backend/db/migrations/20240614204320_build_shards.js
+++ b/apps/backend/db/migrations/20240614204320_build_shards.js
@@ -1,0 +1,59 @@
+/**
+ * @param {import('knex').Knex} knex
+ */
+export const up = async (knex) => {
+  await knex.schema.alterTable("builds", (table) => {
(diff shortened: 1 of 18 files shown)


apps/backend/db/migrations/20240614204320_build_shards.js
```diff
@@ -0,0 +1,59 @@
+/**
+ * @param {import('knex').Knex} knex
+ */
+export const up = async (knex) => {
+  await knex.schema.alterTable("builds", (table) => {
+    table.string("ciProvider");
+    table.string("argosSdk");
```
jsfez, line 7: 🤩

jsfez (approved):

PR #1347, opened 2024-08-12
feat: support gh status for multiple projects linked to same repo

gregberge:


apps/backend/src/build-notification/notification.ts
```diff
@@ -18,8 +20,48 @@ export const NotificationPayloadSchema = z.object({
...
+
+  const query = Project.query();
+
+  if (project.githubRepositoryId) {
+    query.orWhere("githubRepositoryId", project.githubRepositoryId);
+  }
+
+  if (project.gitlabProjectId) {
+    query.orWhere("gitlabProjectId", project.gitlabProjectId);
+  }
```
jsfez, line 39: First one should be "where". Second one should be orWhere.
Maybe it's ok to use "orWhere" for the first one too.

jsfez (approved):

apps/backend/src/build-notification/notification.ts
gregberge: Yeah it's ok, it's better like that because it's aligned.

PR #1351, opened 2024-08-19
fix: avoid concurrent gh commit status creation

gregberge:
Also refactor lock to accept an array

Fix ARGOS-SERVER-XMC

### DIFF
--- a/apps/backend/src/util/redis/lock.ts
+++ b/apps/backend/src/util/redis/lock.ts
@@ -40,11 +40,12 @@ const acquireLock = ({
 
 export const createRedisLock = (client: RedisClient) => {
   async function acquire<T>(
-    name: string,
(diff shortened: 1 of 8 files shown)


apps/backend/src/util/redis/lock.ts
```diff
@@ -40,11 +40,12 @@ const acquireLock = ({
 
 export const createRedisLock = (client: RedisClient) => {
   async function acquire<T>(
-    name: string,
+    key: (string | number)[],
     task: () => Promise<T>,
     { timeout = 20000, retryDelay = { min: 100, max: 200 } } = {},
   ) {
-    const fullName = `lock.${name}`;
+    const hash = key.join(":");
```
jsfez, line 47: Good idea

PR #1370, opened 2024-09-01
fix: comment support multiple projects linked to the same repo

gregberge:


apps/backend/src/build-notification/comment.ts
```diff
@@ -3,42 +3,69 @@ import { invariant } from "@argos/util/invariant";
...
+    .distinctOn("builds.name", "builds.projectId")
     .joinRelated("compareScreenshotBucket")
+    .withGraphFetched("project")
     .where("compareScreenshotBucket.commit", props.commit)
     .orderBy([
-      { column: "name", order: "desc" },
+      { column: "builds.name", order: "desc" },
+      { column: "builds.projectId", order: "desc" },
       { column: "builds.number", order: "desc" },
     ]);
```
jsfez, line 44: To remove

jsfez (approved):

apps/backend/src/build-notification/comment.ts
gregberge: Why, no if it's in distinctOn it's required.

gregberge: If you talk about console log it's done.

PR #1371, opened 2024-09-01
feat: support build metadata

gregberge:
In order to avoid using a build if some tests have failed.


apps/backend/src/build/finalizeBuild.ts
```diff
@@ -1,18 +1,122 @@
...
+
+/**
+ * Check if the bucket is valid from the metadata.
+ */
+export function checkIsBucketValidFromMetadata(metadata: BuildMetadata | null) {
+  return !metadata?.testReport || metadata.testReport.status === "passed";
+}
+
+function add(a: number | undefined, b: number | undefined) {
+  if (!a) {
```
jsfez, line 14: missing `if(!a && !b)` case

apps/backend/src/build/finalizeBuild.ts
```diff
@@ -1,18 +1,122 @@
...
+      if (!acc.testReport?.stats?.startTime) {
+        return item.testReport?.stats?.startTime;
+      }
+      if (!item.testReport?.stats?.startTime) {
+        return acc.testReport.stats.startTime;
+      }
+      return item.testReport.stats.startTime < acc.testReport.stats.startTime
+        ? item.testReport.stats.startTime
+        : acc.testReport.stats.startTime;
+    })();
```
jsfez, line 56:
```suggestion
  return allMetatada.reduce<BuildMetadata | null>((acc, item) => {
    if (!item?.testReport) return acc;
    if (!acc) return item; 
    
    const status = acc.testReport.status === "failed" || item.testReport.status !== "passed"
      ? "failed"
      : "passed";

    const startTime = (() => {
      const accStartTime = acc.testReport?.stats?.startTime;
      const itemStartTime = item.testReport?.stats?.startTime;
      
      if (!accStartTime) return itemStartTime;
      if (!itemStartTime) return accStartTime;
      
      return Math.min(accStartTime, itemStartTime);
    })();
```

apps/backend/src/build/finalizeBuild.ts
gregberge: Not needed.

jsfez (approved):

apps/backend/src/build/finalizeBuild.ts
gregberge: It's 100% equivalent right?

jsfez: Yes, it’s a style improvement suggestion.

PR #1385, opened 2024-10-03
fix: increase api limit

sidharthv96:
E2E test [Failed again](https://github.com/mermaid-js/mermaid/actions/runs/11161194204/job/31023256387). 

I am not sure what the actual size we sent is. 
In case you have logs with this info, please update the size. 

Based on #1380


gregberge (approved):

gregberge:
@sidharthv96 thanks for the PR, let's increase it yeah.

PR #1470, opened 2024-12-15
feat: allow to set a default project user level

gregberge:
CleanShot 2024-12-15 at 11 39 59@2x


apps/frontend/src/router.tsx
```diff
@@ -58,6 +58,7 @@ function RootErrorBoundary() {
   const shouldReload = checkIsFailedToFetchError(error) && !checkHasReloaded();
 
   useEffect(() => {
+    console.error(error);
```
jsfez, line 61: Is this an oversight?

jsfez (approved):

PR #1492, opened 2024-12-31
feat: cache conclusion and stats in database

gregberge:


apps/backend/src/build/job.ts
```diff
@@ -109,8 +97,8 @@ export async function performBuild(build: Build) {
       .withGraphFetched("[gitlabProject, account]")
       .throwIfNotFound(),
     pushBuildNotification({ buildId: build.id, type: "progress" }),
-    createBuildDiffs(build).then(async (diffs) => {
-      pushDiffs(build.id, diffs);
+    createBuildDiffs(build).then(async (screenshotDiffs) => {
+      await pushDiffs({ build, screenshotDiffs });
```
gregberge, line 101: Not good, missing await :/

PR #1495, opened 2025-01-01
refactor: avoid computing stats and conclusion

gregberge:


apps/backend/src/build-notification/comment.ts
```diff
@@ -44,45 +44,45 @@ export const getCommentBody = async (props: {
...
-      // to determine if we should prepend the project name to the build name.
-      .filter((build) => {
-        invariant(build.project, "Relation `project` should be fetched");
-        return build.project.prCommentEnabled;
-      })
-      .map(async (build, index) => {
-        invariant(build.project, "Relation `project` should be fetched");
+  const [aggregateStatuses, urls] = await Promise.all([
+    Build.getAggregatedBuildStatuses(builds),
+    Promise.all(builds.map((build) => build.getUrl())),
```
jsfez, line 49: Why do you need a second `Promise.all` ?

jsfez (approved):

apps/backend/src/build-notification/comment.ts
gregberge: So I can get the urls in an array.

PR #1612, opened 2025-05-06
feat: update copy token button

jsfez:


gregberge (changes requested):
No just remove ARGOS_TOKEN= completely

PR #1635, opened 2025-05-21
feat: send notifications to Slack

### LINKED ISSUE #1531: Slack Integration
- [x] **Multi-channel Support**: Allow configuration of multiple Slack channels for notifications.  
- [x] **Notification Preferences**: Option to receive notifications for **all changes** or only **reference changes**.  
- [x] **Resilient Integration**: Handle scenarios like deleted Slack channels or permission changes without breaking.

jsfez:
Resolve #1531


gregberge (changes requested):
I don't see a problem of notifying in the build-notification job as long as we make the Slack optional (compared to GitLab and GitHub, it's OK to start).

My concern is more about the overall architecture. I will list what's missing here.

## Slack channels handling

The table `slack_channels` is good, it represents a Slack channel. What is missing is the link with `slack_installations` table. A channel has to be linked to a Slack installation and removed before we remove the installation.

We have to sync channels. The UI to add a channel is currently an ID. It's weak I would prefer to have a search in all user channels. But it's acceptable for a v1. However when the user enter the channel we have to resolve it and display the name of the channel. We can't keep the ID because it's not maintainable, a user coming back will not understand what channel is configured.

To do that we have to sync channels. It means we have to fetch the channel from Slack when the user update the channel (to validate it and to get infos). And then we have to listen for events to update the name or eventually to remove the notification and send an email to user if the channel is removed.

## Build Notifications

The table you called `project_slack_channels` is in fact `project_alerts`. It should be something generic to configure the notifications you want to receive when something occurs on a project.

This table has the responsibility to know when we should notify the user. So it means a "notifyOn" very configurable and a "projectId" if it's at project level. In future we may use this table to notify when a test is detected as flaky for example.

But this table should be linked to another table `alert_actions` that will represent the actions to run when this alert is triggered. The first action you develop here is Slack channel, but we will have Webhooks, emails, etc...

So how it should work overall:

- A user link Slack to its account, a Slack installation is created and synced (already working).
- This user creates a new alert, to be notified when a reference build is created for example. He enters the Slack channel as an action of the alert. In the backend, we create and validate the Slack channel, it will be synced properly and we can handle the removal of it, etc.. We also create the alert and the alert action associated to it.
- When a build notification is triggered, we evaluate all active alerts on a project. If one alert matches, we create an "alert_action_delivery" that will take care of the delivery of the alert action asynchronously in a job. In your case, the action is to notify a Slack channel.

PR #1661, opened 2025-06-08
refactor: improve screenshot diff compute

gregberge:


apps/backend/src/screenshot-diff/computeScreenshotDiff.ts
```diff
@@ -226,64 +226,72 @@ export const computeScreenshotDiff = async (
...
+      testId: screenshotDiff.testId,
+      date: new Date(screenshotDiff.createdAt),
+      fileId: diffData.fileId ?? null,
+    });
+  }
+
+  await ScreenshotDiff.query()
+    .findById(screenshotDiff.id)
+    .patch({ ...diffData, jobStatus: "complete" });
+
```
gregberge, line 256: If the job dies here, the build is never concluded (because the jobStatus has been set to "complete" and it's not true, it's not complete.

But we use "jobStatus" in the concludeBuild, so... 

Anyway very few chances that we could have a problem here.

PR #1682, opened 2025-06-13
fix: various fixes

gregberge:


apps/frontend/src/pages/Automation/AutomationFormActionsStep.tsx
```diff
@@ -1,129 +1,130 @@
...
-
-        <Select aria-label="Action Types" onSelectionChange={onSelectionChange}>
+        <Select
+          ref={controller.field.ref}
+          aria-label="Action Types"
+          name={controller.field.name}
+          onBlur={controller.field.onBlur}
+          isDisabled={controller.field.disabled}
+          isInvalid={Boolean(controller.fieldState.error?.message)}
+          selectedKey={null}
```
jsfez: Could be moved in a SelectForm component

gregberge: I did it, I am mostly done for the frontend in my latest commit. The problem is now the GraphQL. I will remove all typing at GraphQL level because it's a mess. Better to rely on JSON schema on the model.

apps/backend/src/graphql/definitions/AutomationRule.ts
```diff
@@ -140,247 +120,184 @@ export const typeDefs = gql`
...
+}) {
+  const { project } = args;
+  const slackChannelActionPayloads = args.input
+    .filter((a) => a.type === "sendSlackMessage")
+    .map((action) =>
+      z
+        .object({
+          slackId: z
+            .string()
+            .max(256, { message: "Must be 256 characters or less" }),
```
jsfez, line 143: The Slack Id I have seen have a length of 11

apps/backend/src/graphql/definitions/AutomationRule.ts
```diff
@@ -140,247 +120,184 @@ export const typeDefs = gql`
...
+          name: z.string().min(1, { message: "Required" }).max(21, {
+            message: "Must be 21 characters or less",
+          }),
```
jsfez: Why the action name can have a maximum length of 21 ?

gregberge: It's the name of the Slack channel but I will remove it we don't care.

apps/backend/src/graphql/definitions/AutomationRule.ts
gregberge: Let's be safe we don't care.

PR #1687, opened 2025-06-17
feat(automations): update submit button order

jsfez:
CleanShot 2025-06-17 at 09 39 28@2x
CleanShot 2025-06-17 at 09 39 17@2x


jsfez:
> The first submit (in the order of the DOM) determines the main action of the form. By changing the order you make "Test notification" the main action of the form. It means when you press "enter" in the name you send a test notification which is not the behaviour expected.

Thank you for the information. I have updated my code accordingly

PR #1748, opened 2025-08-07
feat(builds-page): filter old build names

jsfez:


apps/backend/src/graphql/definitions/Project.ts
```diff
@@ -616,11 +616,23 @@ export const resolvers: IResolvers = {
...
         .distinct("name");
+      if (recentOnly) {
+        // Get builds from the last month
+        const now = new Date();
+        const lastMonth = new Date(
+          now.getFullYear(),
+          now.getMonth() - 1,
+          now.getDate(),
+        );
+        query.where("createdAt", ">=", lastMonth);
```
gregberge, line 633: You can do it in raw sql now() - interval '1 month'

apps/frontend/src/pages/Project/Builds.tsx
```diff
@@ -49,7 +49,7 @@ const ProjectQuery = graphql(`
         id
         url
       }
-      buildNames
+      buildNames(recentOnly: true)
```
gregberge, line 52: Is it used elsewhere without recentOnly? I don't think so, you can probably remove the parameter and only update the resolver?

jsfez: It's used in the automation too

gregberge: You can use the same in automation. Just make sure that the actual value is always part of options.

jsfez: It can be confusing for the code base and for the users. I would prefer to have the whole list in the automation

gregberge: Why would you like to automate a month past build? 🧐

PR #1800, opened 2025-08-31
Sign-up / Sign-in using email

gregberge:
CleanShot 2025-09-06 at 14 41 10@2x
CleanShot 2025-09-06 at 14 41 01@2x
CleanShot 2025-09-06 at 14 40 53@2x
CleanShot 2025-09-06 at 14 40 39@2x


apps/backend/src/util/email.ts
```diff
@@ -1,6 +1,20 @@
...
  */
 export function sanitizeEmail(email: string) {
   return email.trim().toLowerCase();
 }
+
+/**
+ * Get a URL-friendly slug from an email address.
+ */
+export function getSlugFromEmail(email: string): string {
+  const sanitized = sanitizeEmail(email);
```
jsfez, line 14: You can use the Zod's email schema

apps/backend/src/util/request-location.ts
```diff
@@ -0,0 +1,22 @@
...
+}
+
+/**
+ * Extract location information from request headers.
+ */
+export function extractLocationFromRequest(
+  req: Request,
+): RequestLocation | null {
+  const city = req.headers["cf-ipcity"];
+  const country = req.headers["cf-ipcountry"];
```
jsfez, line 15: Why do you request the location? Is it for security purpose?

apps/frontend/src/containers/LoginOptions.tsx
```diff
@@ -1,53 +1,156 @@
...
+              Continue with Google
+            </GoogleLoginButton>
+            <GitHubLoginButton
+              redirect={redirect}
+              size="large"
+              className="w-full justify-center"
+              isDisabled={props.isDisabled}
+            >
+              Continue with GitHub
+            </GitHubLoginButton>
```
jsfez, line 64: Could be cool to have GitHub first

apps/frontend/src/containers/SignupOptions.tsx
```diff
@@ -1,52 +1,175 @@
...
+      >
+        Continue with Google
+      </GoogleLoginButton>
+      <GitHubLoginButton
+        redirect={redirect}
+        size="large"
+        className="w-full justify-center"
+      >
+        Continue with GitHub
+      </GitHubLoginButton>
```
jsfez, line 159: GitHub first please

jsfez (approved):

apps/backend/src/util/email.ts
gregberge: It's not validate but sanitize.

apps/backend/src/util/request-location.ts
gregberge: Yes but it's present in the request because we use Cloudflare proxy.

apps/frontend/src/containers/LoginOptions.tsx
gregberge: It's less friction and less scary to use email or Google to signup. They will have to use GitHub later if they want to import a project. But at the end we will allow to create a project without GitHub. So better to keep this order I think.

PR #1808, opened 2025-09-07
Invitation by email

gregberge:
**Pending invitations**

CleanShot 2025-09-21 at 18 07 17@2x

**New members tab with filters**

CleanShot 2025-09-21 at 18 12 05@2x
CleanShot 2025-09-21 at 18 07 38@2x

**New GitHub members tab**

CleanShot 2025-09-21 at 18 11 36@2x

**New invite dialog**

CleanShot 2025-09-21 at 18 07 26@2x

**Invite screen with auto-join**

CleanShot 2025-09-21 at 18 08 29@2x

**Invitation expired**

CleanShot 2025-09-21 at 18 33 51@2x


apps/backend/db/migrations/20250915191132_team_invites.js
```diff
@@ -0,0 +1,28 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export const up = async (knex) => {
+  await knex.schema.createTable("team_invites", (table) => {
+    table.dateTime("createdAt").notNullable().defaultTo(knex.fn.now());
```
jsfez, line 7: Cool to have default values. I think that the only one table that have it. We should have it on all of createdAt and updatedAt IMO

jsfez, line 7: Cool to have default values. I think that the only one table that have it. We should have it on all of createdAt and updatedAt IMO

jsfez: UpdatedAt column is missing

gregberge: It's never updated so I don't want to add it. Now I keep tables as tiny as possible.

PR #1819, opened 2025-09-14
chore(security): add minimum release age before installing npm packages

gregberge:
https://pnpm.io/blog/releases/10.16


jsfez (commented):
Good !

PR #2103, opened 2026-04-13
feat: improve auth cli success design

jsfez:
CleanShot 2026-04-13 at 18 29 17@2x


gregberge (approved):

gregberge (commented):
I think you could remove lot of text on this page. We don't care where it's saved if we introduce an option to change the directory if will be outdated . And you already said that the auth was successful as title?

PR #2116, opened 2026-04-17
test: add logged build screenshots

jsfez:
## Summary

Adds authenticated Playwright coverage for the existing build page visual scenarios.

The build screenshot matrix now runs only in logged-in mode using `loggedTest`. The authenticated user is attached as an owner of the project team before visiting the build page, and the seeded team account gets a forced Pro-like plan so the captured build pages do not show the subscription banner.

Logged screenshots use the `build-logged-*` naming prefix so Argos can compare authenticated UI states separately from the existing anonymous screenshots.

## Validation

- `pnpm exec playwright test tests/build.spec.ts --project chromium`
- Local result: 14 passed

### DIFF
--- a/tests/build.spec.ts
+++ b/tests/build.spec.ts
@@ -1,8 +1,9 @@
 import { argosScreenshot } from "@argos-ci/playwright";
 import { expect } from "@playwright/test";
 
+import { TeamUser } from "../apps/backend/src/database/models";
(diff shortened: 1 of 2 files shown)


gregberge:
It seems the user is not logged in according to the screenshots?

gregberge (approved):

PR #2128, opened 2026-04-21
feat: support single-domain self-hosted deployments

estevaolucas:
## Problem

Self-hosted Argos instances typically run on a single domain (e.g. `argos.company.com`) without subdomains like `app.argos-ci.com`. Two things break:

1. **Subdomain routing** — the `subdomain` middleware rejects all requests because there is no `app` subdomain, so every route returns a 404.

2. **Auth cookie domain** — the JWT cookie is set with domain `.argos-ci.com` in production, which doesn't match the self-hosted domain. Authentication silently fails because the browser never sends the cookie back.

## Solution

Uses the existing `selfHosted` config flag (from #2114):

- **Backend**: bypass subdomain matching when `selfHosted` is `true` (`util.ts`)
- **Frontend**: leave cookie domain empty (defaults to current host) when `selfHosted` is `true` (`Auth.tsx`)
- **Config**: expose `selfHosted` in `ClientConfig` so the frontend can read it from `/config.js` (`config-types`, `app-router.ts`)

No breaking changes — the `selfHosted` flag defaults to `false`, so existing multi-subdomain deployments are unaffected.


apps/frontend/src/containers/Auth.tsx
```diff
@@ -23,8 +23,14 @@ interface AuthContextValue {
 const AuthContext = createContext<AuthContextValue | null>(null);
 
 const COOKIE_NAME = "argos_jwt";
-const COOKIE_DOMAIN =
-  process.env["NODE_ENV"] === "production" ? ".argos-ci.com" : "";
+const COOKIE_DOMAIN = (() => {
+  // Self-hosted deployments use a single domain; no cross-subdomain cookie needed.
+  const clientConfig = (window as any).clientData?.config;
```
gregberge: Use apps/frontend/src/config.ts

gregberge (approved):

estevaolucas:
@gregberge thanks for the review. Any recommendation on what to do with those failing CI jobs, which doesn't seem related to my PR?

gregberge:
@estevaolucas same as the other, rebase and fix CI please

PR #2130, opened 2026-04-22
fix(notifications): fix usage based notifications

jsfez:


apps/backend/src/database/models/User.ts
```diff
@@ -142,4 +147,18 @@ export class User extends Model {
...
+  static getNotificationEmailAddress(
+    user: NotificationEmailCandidate,
+  ): string | null {
+    if (user.email) {
+      return user.email;
+    }
+
+    const verifiedEmail = user.emails?.find((email) => email.verified);
+    return verifiedEmail?.email ?? null;
+  }
```
gregberge, line 163: It's useless because a user always has a primary email and it should be verified. For notifications we only use the primary email so you can remove this helper.

If we have some user with verified emails and not a primary email then it's the source of the problem and it's where we need to fix it.

apps/backend/src/database/services/spend-limit.ts
```diff
@@ -48,8 +48,8 @@ export async function getSpendLimitThreshold(input: {
       if (
         // The highest threshold is reached.
         (acc === null || acc < threshold) &&
-        (previousUsageCost === null || previousUsageCost <= limitAtThreshold) &&
-        currentCost > limitAtThreshold
+        (previousUsageCost === null || previousUsageCost < limitAtThreshold) &&
+        currentCost >= limitAtThreshold
```
gregberge, line 52: For me the previous code is equivalent. It was notifying when the limit was crossed now it's when the limit is reached. But it was OK.

apps/backend/src/database/services/spend-limit.ts
```diff
@@ -75,5 +75,5 @@ export async function checkIsBlockedBySpendLimit(
   }
 
   const currentCost = await manager.getAdditionalScreenshotCost();
-  return currentCost > account.meteredSpendLimitByPeriod;
+  return currentCost >= account.meteredSpendLimitByPeriod;
```
gregberge, line 78: Also similar.

gregberge (changes requested):
For me you changed things but you didn't fix the bug. The logic on emails is not good.

apps/backend/src/notification/handlers/spend_limit.tsx
```diff
@@ -80,6 +80,19 @@ export const handler = defineNotificationHandler({
...
+                        team reaches <strong>100%</strong> of the spend limit.
+                      </Paragraph>
+                    ) : (
+                      <Paragraph>
+                        Automatic pause is disabled, so{" "}
+                        <strong>builds will continue</strong> even if your team
+                        goes above <strong>100%</strong> of the spend limit.
+                      </Paragraph>
+                    )}
                   </>
```
gregberge, line 96: I think it's clear enough

PR #2134, opened 2026-04-24
Cast githubPullRequestId to string

th1nkful:
The database expects this to be a string. A few lines above, `pullRequest.id` is cast to a string but not here. This was causing validation failures from the CLI `upload` command when running inside a `merge_group` job.

Happy for you to just close this but thought it might be valuable to pass it along.

I have [Argos CI running in Cloud Run](https://github.com/argos-ci/argos/discussions/1811) and was having trouble with GitHub Actions in `merge_group` jobs being unable to upload to Argos despite including the token and even trying to pass it myself. It might be a configuration issue but I can't see how as it is being returned from Knex/postgres.

### DIFF
--- a/apps/backend/src/build/createBuild.ts
+++ b/apps/backend/src/build/createBuild.ts
@@ -191,7 +191,7 @@ export async function createBuild(params: {
           await BuildMergeQueueGhPullRequest.query(trx).insert(
             mergeQueuePullRequests.map((pullRequest) => ({
               buildId: build.id,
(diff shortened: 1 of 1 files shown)


gregberge:
Hi @th1nkful the fix is most likely invalid. The id is a biginteger and treated as string in the codebase. It may be a problem in your self-hosted instance, but as we said we don't provide support for it (except if you pay for it). Since this bug is not in the Cloud version of Argos and that at typing level there is no problem we won't merge it.

PR #2135, opened 2026-04-24
fix: add periodic reconciliation for builds stuck without conclusion

estevaolucas:
## Problem

Builds can finish uploading (`jobStatus=complete`) but never get a `conclusion` set, leaving GitHub status checks stuck at "Build in progress..." indefinitely.

Root causes observed in self-hosted deployments:
- **Redis lock contention**: `concludeBuild` uses a Redis lock. When 1,000+ diffs complete near-simultaneously, many lock acquisitions time out while the actual last diff is being processed — if the lock holder sees any diff still `pending` at that moment, it returns null without concluding
- **Process crash**: worker crashes after setting diff `jobStatus=complete` but before `concludeBuild` fires
- **SQS message dropped**: the final diff message is never processed

**Observed in production**: 47 builds stuck in a single day with `conclusion=null` despite `jobStatus=complete` and all diffs computed.

## Solution

Add `reconcileStaleBuilds()` — runs every 5 minutes via the existing `scheduleCron` mechanism in the `build-and-synchronize` process.

Logic:
1. Find builds with `jobStatus=complete`, `conclusion=null`, finalized > 2 minutes ago
2. Filter to only builds where no diffs remain in `pending` state
3. Re-run `concludeBuild()` for each — the existing lock prevents double-conclude

The 2-minute grace period avoids racing with in-flight diff processing for very new builds.

## Testing

Manually verified by setting `conclusion=null` on a completed build and confirming the reconciler re-sets it correctly.

### DIFF
--- a/apps/backend/src/build/reconcileStaleBuilds.ts
+++ b/apps/backend/src/build/reconcileStaleBuilds.ts
@@ -1,0 +1,55 @@
+import { Build, ScreenshotDiff } from "@/database/models";
+
+import { concludeBuild } from "./concludeBuild";
+
+/**
+ * Find builds that are complete but never got a conclusion set.
+ *
+ * This can happen when:
+ * - The Redis lock for concludeBuild times out while 1000s of diffs complete simultaneously
+ * - The process crashes after setting jobStatus=complete but before concludeBuild fires
+ * - The SQS message for the final diff is dropped
+ *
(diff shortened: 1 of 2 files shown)


gregberge (changes requested):
This is not the source of the problem, if builds are not processed correctly then we have to find a way to process them correctly. On Argos public instance we don't need this. So I am not in favor of it.

PR #2139, opened 2026-04-28
fix: support CI on fork PRs

estevaolucas:
## Summary
- Split CI into two workflows to support fork PRs:
  - **ci.yml** (`pull_request`) — static-analysis, unit-test, integration-test
  - **ci-e2e.yml** (`pull_request_target`) — storybook-test, storybook-publish, e2e-test
- Jobs that require secrets (Argos tokens, AWS credentials) use `pull_request_target` so fork PRs get access to them
- All checkout steps in `ci-e2e.yml` use explicit `ref: ${{ github.event.pull_request.head.sha || github.sha }}` to test the actual PR code

## Why split instead of using `pull_request_target` everywhere?
`pull_request_target` exposes repository secrets to fork PR code. Limiting it to only the jobs that need secrets reduces the attack surface — a malicious fork PR could potentially exfiltrate secrets if all jobs ran with them. By keeping static-analysis, unit-test, and integration-test on `pull_request`, those jobs never see secrets on fork PRs.

Requires "Require approval for first-time contributors" in repo settings so a maintainer reviews fork PRs before CI triggers.

### DIFF
--- a/.github/workflows/ci-e2e.yml
+++ b/.github/workflows/ci-e2e.yml
@@ -1,0 +1,142 @@
+name: CI (E2E & Storybook)
+permissions: {}
+
+on:
+  push:
+    branches: [main]
+  pull_request_target:
+    branches: [main]
+
+env:
+  TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
+  TURBO_TEAM: ${{ vars.TURBO_TEAM }}
+  SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+
+jobs:
+  storybook-test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    env:
+      NODE_ENV: test
+      TZ: utc
+      ARGOS_TOKEN: ${{ secrets.ARGOS_STORYBOOK_TOKEN }}
+
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v6
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+
+      - name: Setup Node.js & install dependencies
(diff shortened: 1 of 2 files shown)


gregberge:
Is it possible to run E2E + Storybook tests + Storybook publish on `pull_request_target` event instead?

estevaolucas:
Switched to `pull_request_target` — all jobs now run with secrets on fork PRs. Each checkout step explicitly uses `ref: ${{ github.event.pull_request.head.sha || github.sha }}` to test the actual PR code (since `pull_request_target` defaults to the base branch).

The `describe.skipIf` guards on S3 tests are kept as a defensive measure for local dev without credentials, but they shouldn't be needed in CI anymore.

Note: `pull_request_target` does mean fork PR code runs with secrets. This is safe as long as "Require approval for first-time contributors" is enabled (so a maintainer reviews before CI triggers).

estevaolucas:
Stripped the PR down to just the workflow change as you suggested — replaced `pull_request` with `pull_request_target` and added explicit `ref: ${{ github.event.pull_request.head.sha || github.sha }}` to all checkout steps. No test file changes needed since all jobs now get secrets.

Note: this change can't be validated until merged, since GitHub reads the workflow file from the base branch.

estevaolucas:
Updated the approach — split CI into two workflow files:

- **ci.yml** stays on `pull_request` for static-analysis, unit-test, integration-test
- **ci-e2e.yml** uses `pull_request_target` for storybook-test, storybook-publish, e2e-test (the jobs that need secrets)

We opted for the split instead of applying `pull_request_target` to everything because it exposes repository secrets to fork PR code. Keeping it scoped to only the jobs that actually need secrets minimizes the attack surface. Let us know if you think there's a better approach.

PR #2140, opened 2026-04-29
fix: clamp self-hosted includedScreenshots to GraphQL Int max

estevaolucas:
## Problem

Self-hosted instances need to bypass Stripe billing and return unlimited screenshot counts. The bypass uses Number.MAX_SAFE_INTEGER (9007199254740991) as the sentinel for unlimited, but GraphQL Int is a 32-bit signed integer (max 2,147,483,647). This causes INTERNAL_SERVER_ERROR and nullifies the account field, making the settings page return NotFound.

## Context on selfHosted config

PR #2114 added a general selfHosted config flag (env: SELF_HOSTED). PR #2128 replaced it with the narrower trackNpmPackagesVersions (env: TRACK_NPM_PACKAGE_VERSIONS) to scope it specifically to npm registry behavior.

This PR proposes adding selfHosted back as a general flag — the billing bypass needs to gate subscription manager behavior broadly, which goes beyond npm registry tracking. Happy to discuss alternatives (e.g. forcedPlanId-based detection).

## Fix

- Add selfHosted: boolean to config (env: SELF_HOSTED)
- Use config.get('selfHosted') not process.env directly, per project conventions
- Add self-hosted subscription manager returning enterprise-equivalent values
- Use 2_147_483_647 (max 32-bit int) as the unlimited sentinel instead of Number.MAX_SAFE_INTEGER

## Testing

Navigate to account settings on a self-hosted instance. Without this fix it returns NotFound; with it settings load correctly.

### DIFF
--- a/apps/backend/src/config/index.ts
+++ b/apps/backend/src/config/index.ts
@@ -73,6 +73,12 @@ export function createConfig() {
       default: true,
       env: "TRACK_NPM_PACKAGE_VERSIONS",
     },
+    selfHosted: {
+      doc: "Whether this is a self-hosted instance. Disables cloud-only features (billing, npm version tracking, etc.).",
+      format: Boolean,
+      default: false,
+      env: "SELF_HOSTED",
+    },
     samlTeamSlug: {
       doc: "The team account slug to auto-redirect to on login. Enables SSO-only login flow. Useful only for self-hosted.",
       format: String,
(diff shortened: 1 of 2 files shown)


gregberge (changes requested):
No this is ugly, we already have a plan logic, just use it.

PR #2141, opened 2026-04-29
fix: skip app router catch-all for /v2 in single-domain self-hosted mode

estevaolucas:
## Problem

In single-domain self-hosted deployments (using `WILDCARD_DOMAINS` from #2128), all requests are routed to the app router. The app router has a catch-all `/{*splat}` that serves `index.html` for any unmatched path.

This means `GET /v2/project`, `POST /v2/builds`, etc. all return `index.html` instead of API responses, completely breaking CI integrations on self-hosted instances.

## Fix

Add a `/v2` handler early in the app router that calls `next('router')` when `SELF_HOSTED=true`. This skips the app router and passes the request to the API router registered next in `app.ts`.

## Context

This pairs with `wildcardDomains` support added in #2128. Without this fix, single-domain self-hosted instances cannot accept any API calls from CI.

### DIFF
--- a/apps/backend/src/config/index.ts
+++ b/apps/backend/src/config/index.ts
@@ -73,6 +73,12 @@ export function createConfig() {
       default: true,
       env: "TRACK_NPM_PACKAGE_VERSIONS",
     },
+    selfHosted: {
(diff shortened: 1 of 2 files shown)


gregberge (changes requested):
I understand the concern, so it becomes messy, I don't know why using one single domain for API and app should be useful.

PR #2142, opened 2026-04-29
fix: add periodic reconciliation for builds stuck without conclusion

estevaolucas:
## Problem

Builds can get stuck in a `pending` jobStatus with no `conclusion` indefinitely when:
- The worker processing the build crashes mid-run
- The worker pod is restarted (e.g. during a deploy)
- A transient infrastructure error occurs

These stuck builds block GitHub CI checks forever, requiring manual intervention to unblock PRs.

## Fix

Add a periodic reconciliation task (every 10 minutes) that finds builds with `jobStatus='pending'` older than 30 minutes that have no conclusion, then forces them to a `failure` conclusion. This unblocks GitHub checks automatically.

Note: This was previously submitted as #2135 (closed). Resubmitting with the same approach.

### DIFF
--- a/apps/backend/src/config/index.ts
+++ b/apps/backend/src/config/index.ts
@@ -73,6 +73,12 @@ export function createConfig() {
       default: true,
       env: "TRACK_NPM_PACKAGE_VERSIONS",
     },
+    selfHosted: {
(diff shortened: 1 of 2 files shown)


gregberge:
I already said no to this one.

PR #2143, opened 2026-04-29
fix: expose selfHosted flag and nullable samlTeamSlug in ClientConfig

estevaolucas:
## Problem

Self-hosted instances need to communicate their mode to the frontend so it can:
- Auto-redirect the login page to the configured SSO provider
- Hide cloud-only features (billing, marketplace links, etc.)

Currently there's no way for frontend code to know it's running on a self-hosted instance.

Also, `samlTeamSlug` is typed as `string` but is `null` on cloud instances where SSO isn't configured — the frontend needs to check for null before using it.

## Fix

- Add `selfHosted: boolean` to `ClientConfig`, populated from `process.env["SELF_HOSTED"] === "true"` in `app-router.ts`
- Change `samlTeamSlug: string` to `samlTeamSlug: string | null` to accurately reflect that it's optional

## Usage

```ts
if (config.selfHosted && config.samlTeamSlug) {
  // Auto-redirect to SSO on login page
  redirectToSAMLLogin({ teamSlug: config.samlTeamSlug });
}
```

### DIFF
--- a/apps/backend/src/config/index.ts
+++ b/apps/backend/src/config/index.ts
@@ -73,6 +73,12 @@ export function createConfig() {
       default: true,
       env: "TRACK_NPM_PACKAGE_VERSIONS",
     },
+    selfHosted: {
+      doc: "Whether this is a self-hosted instance. Disables cloud-only features (billing, npm version tracking, etc.).",
+      format: Boolean,
+      default: false,
+      env: "SELF_HOSTED",
+    },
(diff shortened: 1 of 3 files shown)


gregberge:
We don't add self-hosted property and change the behaviour, instead we will add some configuration to make the correct behaviour. You assume it's what a user want in self-hosted but it's not true, self-hosted is just your custom config.

PR #2220, opened 2026-05-26
feat: support new API headers

jsfez:


apps/backend/src/api/handlers/updateBuild.ts
```diff
@@ -121,13 +135,25 @@ type Context = {
...
   project: Project;
   build: Build;
+  requestId: string | null;
 };
 
+async function hasShardBeenProcessed(params: {
+  buildId: string;
+  requestId: string;
+}) {
+  const shard = await BuildShard.query().findOne({
```
gregberge:
```suggestion
  const shard = await BuildShard.query().select('id').findOne({
```

apps/backend/src/api/handlers/updateBuild.ts
```diff
@@ -141,21 +167,28 @@ async function handleUpdateParallel(ctx: Context) {
...
+            return false;
+          }
+        }
+
+        const shard = await BuildShard.query(trx).insert({
+          buildId: build.id,
+          index: body.parallelIndex ?? null,
+          nonce: requestId,
+          metadata: body.metadata ?? null,
+        });
```
gregberge: This is not the good way to do it, we should insert and if there is a conflict we handle it. Here we can have some concurrency issue. See other `onConflict` we use in the app.

apps/backend/src/web/app.ts
```diff
@@ -49,6 +49,18 @@ export const createApp = async (): Promise<express.Express> => {
...
+        "clientRetryAttempt",
+        req.headers["x-argos-retry-attempt"] as string,
+      );
+    }
+    if (req.headers["x-argos-request-id"]) {
+      scope.setTag(
+        "clientRequestId",
+        req.headers["x-argos-request-id"] as string,
+      );
+    }
```
gregberge: - These headers are specific to the API you are at the root of the app here, it's not the good place.
- The scope is not used for logger, that said it's good to have it here but I would like to also have it in the logs.

gregberge (changes requested):

gregberge (approved):

PR #2227, opened 2026-05-28
feat: support arbitrary text-based snapshot types

gregberge:
## Summary
- Add support for text-based snapshots (JSON, YAML, XML, HTML, markdown, plain text, …) alongside the existing image and ARIA snapshots, gated by a shared content-type allowlist in `@argos/schemas/content-type`. Server-side, the input content type is validated against the allowlist and normalized (lowercased, parameters stripped).
- Harden non-image file serving: override the presigned GET with `Content-Disposition: attachment` + `Content-Type: text/plain; charset=utf-8` so attacker-controlled content (e.g. HTML, SVG) cannot be rendered as active content on the storage origin. In-app rendering of text snapshots is unchanged (Monaco source view).
- Move the upload size limit into the signed URL itself: switch `createBuild` from presigned PUT to `createPresignedPost` with a `["content-length-range", 1, 25 MB]` condition. S3 itself rejects oversized uploads before they hit the bucket — no server-side HEAD or post-hoc validation needed.
- Frontend Monaco language map gains `application/xml` and normalizes the content type (handles `; charset=…` etc.). The diff downloader falls back to a generic extension when `mime` cannot resolve one (e.g. `application/yaml`).
- Documents how to run backend tests in `apps/backend/AGENTS.md`.

## Breaking change (API + SDK)
The `createBuild` response changed from `{ key, putUrl }` to `{ key, url, fields }` (S3 POST policy). The Argos SDK upload now needs to send a multipart `POST` to `url` with `fields` merged into the form and the file part appended last. **The SDK update lands in a follow-up PR in the SDK repo** — this PR is API + UI only.

## Test plan
- [x] Unit — `ScreenshotInputSchema`: defaults, allowlist acceptance, normalization, rejection of unsupported / unsafe content types (incl. `image/svg+xml`).
- [x] Unit — `getSignedObjectUrl`: response-content-disposition / response-content-type query params for downloads, absent by default.
- [x] e2e — `snapshot-diff` serialization still returns correct image URLs.
- [x] e2e — `api-v2` full create-build → multipart POST upload → update-build flow against real S3 (7/7).
- [x] Backend / frontend / schemas typecheck, lint, prettier.
- [x] SDK update: https://github.com/argos-ci/argos-javascript/pull/313

### DIFF
--- a/apps/backend/AGENTS.md
+++ b/apps/backend/AGENTS.md
@@ -52,3 +52,26 @@
 - Avoid large shared setups that slow down the suite or reduce concurrency.
 - Load dependencies independently whenever possible so tests can run in parallel without unnecessary bottlenecks.
 - Avoid mocking unless the behavior cannot be tested realistically.
+
+### Running tests
+
+From the repo root:
+
+```bash
+pnpm test:unit          # unit tests (no infra)
+pnpm test:integration   # e2e tests (require Postgres + Redis)
+```
+
+Before running e2e tests for the first time — and after pulling new migrations — reset the test database so its schema matches `db/structure.sql`:
+
+```bash
+NODE_ENV=test pnpm run --filter @argos/backend db:reset
+```
+
+If `db:reset` reports `database "test" is being accessed by other users`, terminate the open sessions first:
+
+```bash
+psql -h 127.0.0.1 -U postgres -d postgres -c \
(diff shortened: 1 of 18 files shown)


jsfez (commented):
lgtm

PR #2240, opened 2026-05-30
feat(comment): limit the size of a comment

gregberge:
## What

Prevents users from posting very large comments by enforcing size limits on both the client and the server.

### Client character limit
- [Editor.tsx](apps/frontend/src/ui/Editor/Editor.tsx) loads TipTap's `CharacterCount` extension, capping input at **2,000 characters**. No UI — it just silently prevents typing past the cap.

### Server validation
- [validate.ts](apps/backend/src/comment/validate.ts) gains `isCommentTooLarge()`, which walks the document once and rejects it if it exceeds any of:

  | Limit | Value | Protects |
  |---|---|---|
  | characters | 2,000 | UX / walls of text |
  | JSON bytes | 20,000 | storage & bandwidth abuse |
  | nodes | 300 | rendering |
  | depth | 12 | recursive rendering / pathological nesting |

  The schema/node-mark whitelist was already enforced by `validateCommentJson`, so it isn't duplicated.

### Wired into all comment entry points
- `createBuildComment`, `updateBuildComment`, `createBuildReview` — `throw boom(400, "Comment is too large")`.
- `createReview` API handler — added a Zod `.refine` with a "Comment is too large." message.

## Notes
The client cap and server `MAX_COMMENT_CHARACTERS` are defined independently in the two packages (no shared constant) but kept in sync.

## Tests
Added 6 unit tests to [validate.test.ts](apps/backend/src/comment/validate.test.ts) covering each limit. All 21 unit tests pass; `tsc` and `eslint` clean.

### DIFF
--- a/apps/backend/src/api/handlers/createReview.ts
+++ b/apps/backend/src/api/handlers/createReview.ts
@@ -12,7 +12,7 @@ import {
   createBuildReview,
   ScreenshotDiffReviewState,
 } from "@/build/createBuildReview";
-import { validateCommentJson } from "@/comment/validate";
+import { isCommentTooLarge, validateCommentJson } from "@/comment/validate";
 import { Build } from "@/database/models/Build";
 import { boom } from "@/util/error";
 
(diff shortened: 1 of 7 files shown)


jsfez (commented):
LGTM

PR #2250, opened 2026-06-08
feat: Implement account lockout mechanism for brute-force protection

jsfez:
## Summary

Implements a robust account lockout mechanism to prevent brute-force attacks on email-based authentication, addressing the critical security vulnerability reported in ARG-384.

## Problem

The application previously allowed unlimited authentication attempts, making it vulnerable to brute-force attacks. Attackers could systematically guess the 6-digit login code with no rate limiting or account lockout mechanism.

## Solution

The implementation adds per-account rate limiting with automatic account lockout:

- **5 failed attempts** trigger a **30-minute account lockout**
- Failed attempt counters expire after 30 minutes of inactivity
- Successful authentication clears all failed attempts
- HTTP 429 (Too Many Requests) returned when locked, with friendly messaging
- Uses Redis for fast, atomic operations

## Key Changes

**`apps/backend/src/auth/email.ts`** - Core lockout logic
- Enhanced `verifyAuthEmailCode()` to track failed attempts and enforce lockout
- Returns structured result: `{ valid, locked, remainingTime }`
- Automatically locks accounts after 5 failed attempts
- Provides remaining lockout time to caller

**`apps/backend/src/database/services/account.ts`** - Authentication flow
- Updated `authenticateWithEmail()` to handle lockout responses
- Returns HTTP 429 with user-friendly error message showing retry time
- Separates locked vs. invalid code errors

**`apps/backend/src/graphql/definitions/Account.ts`** - GraphQL error handling
- Handles both `ACCOUNT_LOCKED` and `INVALID_EMAIL_VERIFICATION_CODE` errors
- Returns appropriate GraphQL error responses

**`packages/error-types/types.d.ts`** - Error types
- Added `ACCOUNT_LOCKED` error code for type safety

## Security Impact

- Limits attackers to 5 attempts per email per 30-minute window
- Makes brute-force attacks computationally infeasible
- Maintains user experience for legitimate users (auto-unlock after 30 min)
- Prevents repeated lockouts by clearing attempts on successful login

Closes #ARG-384


apps/backend/src/auth/email.ts
```diff
@@ -20,20 +32,114 @@ export async function generateAuthEmailCode(email: string) {
...
+  valid: boolean;
+  locked: boolean;
+  remainingTime?: number;
+}> {
   const { code, email } = args;
   const redis = await getRedisClient();
+
+  // Check if account is locked
+  if (await isAccountLocked(email)) {
+    const remainingTime = await getRemainingLockoutTime(email);
```
gregberge: It should be the same, by checking the TTL you can know if the key exists are not, so it's useless to do two calls here.

apps/backend/src/auth/email.ts
```diff
@@ -20,20 +32,114 @@ export async function generateAuthEmailCode(email: string) {
...
+    };
+  }
+
   const key = getRedisKey(email);
   const storedCode = await redis.get(key);
+
   if (storedCode !== String(code)) {
-    return false;
+    // Track failed attempt
+    const attempts = await incrementFailedAttempts(email);
```
gregberge: Your implementation does not work, you are locking only when there is a fail. So it means if someone sends 10000 requests at the same time, one of them could be valid.

The overall algorithm is not safe. It should be a kind of lock that increase a number and in response check if there number is below a thresold of attempts.

gregberge (changes requested):

gregberge (approved):

PR #2273, opened 2026-06-16
feat: anchor comments to screenshots and snapshots

gregberge:
## What

Adds **contextual comments**: a comment can now reference a specific screenshot diff, optionally pinned to a normalized point on one side (baseline/compare) or a line range on a textual snapshot. This is phase 1 (backend + sidebar display & creation); on-canvas pins, the Figma-style comment tool, and snapshot-line creation UI are deferred to a follow-up (the backend already supports all three anchor kinds).

The anchor lives on the **root comment** only — replies inherit it via their thread, like `resolvedAt`.

## Backend

- **Migration** adds `screenshotDiffId` (FK → `screenshot_diffs`, indexed, `ON DELETE CASCADE`) and `anchor` (jsonb) to `comments`, with a CHECK constraint that an anchor can't exist without a diff.
- **Comment model** gains the fields, the `screenshotDiff` relation, and a `CommentAnchor` discriminated-union type (`point` with `side`/`x`/`y`, or `lines` with `from`/`to`).
- **`comment/anchor.ts`** validates bounds (coords in 0–1; positive, non-inverted line ranges).
- **`createBuildComment`** persists `screenshotDiffId` + `anchor`.
- **GraphQL**: `Comment.screenshotDiff` + `Comment.anchor: CommentAnchor` (union of `CommentPointAnchor`/`CommentLinesAnchor`); `AddBuildCommentInput` gains `screenshotDiffId` + `anchor`. The resolver validates the diff belongs to the build, requires a diff when an anchor is set, and rejects anchors on replies.

## Frontend

- **`CommentScreenshotReference`**: anchored comments render a clickable snapshot reference (thumbnail + name, pin icon for points, "Lines X–Y" for line anchors) that jumps to the diff in the viewer.
- **Composer**: attaches the snapshot currently in view by default, shown as a thumbnail toggle next to the send button — click to detach (shows an eye-off icon) or re-attach. Added a `footerStart` slot to `StandaloneEditor` to host it without adding vertical space.

## Testing

- 17 e2e tests in `addBuildComment.e2e.test.ts` cover whole-diff / point / lines anchors plus rejection of reply-anchors, cross-build diffs, anchor-without-diff, out-of-range coords, and both-anchors-set.
- Typecheck, lint, and knip pass.


apps/backend/src/database/models/Comment.ts
```diff
@@ -26,6 +54,8 @@ export class Comment extends Model {
           buildId: { type: "string" },
           buildReviewId: { type: ["string", "null"] },
           threadId: { type: ["string", "null"] },
+          screenshotDiffId: { type: ["string", "null"] },
+          anchor: { type: ["object", "null"] },
```
gregberge: Use a zod type and convert it to JSON Schema + infer the type so it's consistent.

apps/backend/src/comment/anchor.ts
```diff
@@ -0,0 +1,53 @@
...
+      }
+      if (anchor.to < anchor.from) {
+        throw boom(400, "Comment anchor line range is inverted");
+      }
+      return;
+    }
+    default:
+      throw boom(400, "Invalid comment anchor type");
+  }
+}
```
gregberge: All of this should be done in the Zod schema of the anchor that we should use in the model.

apps/backend/src/graphql/definitions/Comment.ts
```diff
@@ -144,6 +177,34 @@ export const typeDefs = gql`
     users: [User!]!
   }
 
+  "Which side of a screenshot diff a position anchor points at."
+  enum CommentAnchorSide {
+    baseline
+    compare
+  }
```
gregberge: I am not sure the side is useful because the two sides are displayed at the exact same scale so the coordinates on one or another are the same.

apps/backend/src/graphql/definitions/Comment.ts
```diff
@@ -379,11 +493,39 @@ export const resolvers: IResolvers = {
...
+      let screenshotDiffId: string | null = null;
+      if (input.screenshotDiffId) {
+        const diff = await ScreenshotDiff.query().findById(
+          input.screenshotDiffId,
+        );
+        if (!diff || diff.buildId !== build.id) {
+          throw notFound("Screenshot diff not found");
+        }
+        screenshotDiffId = diff.id;
+      }
```
gregberge, line 513: We don't really have to validate this, it will not work if the idea does not exist anyway.

gregberge: But at some point we should validate that this diff is part of the build.

PR #2287, opened 2026-06-21
feat(api): expose review and commenting actions in the REST API

gregberge:
## What

Exposes the full review + commenting surface — previously available only in GraphQL — through the public **v2 REST API**, so it can be driven from the `@argos-ci/cli`. Only `POST .../reviews` (`createReview`) existed before.

### New endpoints (nested under the build path, PAT-only)

**Reviews**
- `GET  .../builds/{n}/reviews` — list reviews (submitted + the caller's own draft)
- `POST .../builds/{n}/reviews/{reviewId}/dismiss`
- `createReview` body extended to accept Markdown

**Comments**
- `GET  .../comments` · `POST .../comments` — list / create (reply via `threadId`, anchor to a diff, `addToReview` draft)
- `GET .../comments/{id}` · `PATCH .../comments/{id}` · `DELETE .../comments/{id}`
- `POST .../comments/{id}/reactions` · `DELETE .../comments/{id}/reactions?emoji=…`
- `POST .../comments/{id}/resolve` · `POST .../comments/{id}/unresolve`
- `POST .../comments/{id}/subscription` · `DELETE .../comments/{id}/subscription`

## Design

- **Body = Markdown _or_ rich-text JSON.** `resolveCommentBody` converts Markdown → HTML (`marked`) → ProseMirror JSON via the existing comment schema; objects are validated as raw JSON. `createReview`'s `body` now accepts both, backward-compatibly. (Limitation: `@mentions` need user ids and aren't expressible in Markdown — use the raw-JSON form.)
- **Reuses GraphQL services, no duplication.** Handlers call the same service functions; shared logic — review dismissal, comment visibility, thread resolution, reviewable-status — is extracted into modules consumed by both REST and the GraphQL resolvers/loader, preserving GraphQL's typed errors.
- **Conventions:** PAT-only (these are user actions and pending-draft visibility depends on the viewer); raw DB ids like the rest of v2; draft comments on pending reviews stay visible only to their author; deleted comments hidden. Comments serialize both the rich-text `body` and a plain-text `text` (batched to avoid N+1).
- Added `patch`/`delete` support to the API handler layer; new serializers and a shared build/comment auth helper.

The OpenAPI spec (`/v2/openapi.yaml`) regenerates automatically from `schema.ts`.

## Testing

- New e2e suites cover every handler: Markdown/raw-JSON bodies, replies, anchors, `addToReview` drafts, reactions, resolve/unresolve, subscriptions, and auth/permission/visibility failure paths (401/403/404/400).
- Full backend suite green for the touched areas (283 e2e + 81 unit, incl. 26 new). Typecheck + lint clean.

### DIFF
--- a/apps/backend/src/api/handlers/createComment.ts
+++ b/apps/backend/src/api/handlers/createComment.ts
@@ -1,0 +1,214 @@
+import { z } from "zod";
+import { ZodOpenApiOperationObject } from "zod-openapi";
+
+import { getOrCreatePendingBuildReview } from "@/build/pendingReview";
+import { isReviewableBuildStatus } from "@/build/reviewableStatus";
+import { resolveCommentBody } from "@/comment/body";
+import { createBuildComment } from "@/comment/createBuildComment";
+import { Build, BuildReview, ScreenshotDiff } from "@/database/models";
+import {
+  CommentAnchorSchema,
+  type CommentAnchor,
+} from "@/database/models/Comment";
+import { boom } from "@/util/error";
+
+import {
+  assertBuildPermission,
+  getBuildCommentThread,
+  getPatAuthAndBuild,
+} from "../auth/build";
+import { BuildNumber } from "../schema/primitives/build";
+import {
+  CommentBodyInputSchema,
+  CommentSchema,
+  serializeComment,
+} from "../schema/primitives/comment";
+import { AccountSlug, ProjectName } from "../schema/primitives/project";
+import {
+  forbidden,
+  invalidParameters,
+  notFound,
+  serverError,
+  unauthorized,
+} from "../schema/util/error";
+import { CreateAPIHandler } from "../util";
+
+const AnchorInputSchema = z
+  .object({
+    point: z.object({ x: z.number(), y: z.number() }).optional().meta({
+      description: "A point on the diff, in normalized (0–1) coordinates.",
+    }),
+    lines: z
+      .object({ from: z.number().int(), to: z.number().int() })
+      .optional()
+      .meta({ description: "A 1-based inclusive line range on a text diff." }),
+  })
+  .meta({
+    description:
+      "Where on the screenshot diff the comment points. Provide exactly one of point or lines.",
+  });
+
+const CreateCommentBodySchema = z.object({
+  body: CommentBodyInputSchema,
+  threadId: z
+    .string()
+    .optional()
+    .meta({ description: "Root comment ID to reply to." }),
+  screenshotDiffId: z.string().optional().meta({
+    description:
+      "Screenshot diff to anchor the comment to. Required when anchor is set.",
+  }),
+  anchor: AnchorInputSchema.optional(),
+  addToReview: z.boolean().optional().meta({
+    description:
+      "Attach the comment to your pending review (created if needed) instead of posting it immediately. Ignored for replies, which inherit their thread's review.",
+  }),
+});
+
+export const createCommentOperation = {
+  operationId: "createComment",
+  summary: "Post a comment (or reply) on a build",
+  requestParams: {
+    path: z.object({
+      owner: AccountSlug,
+      project: ProjectName,
+      buildNumber: BuildNumber,
+    }),
+  },
+  requestBody: {
+    required: true,
+    content: {
+      "application/json": {
+        schema: CreateCommentBodySchema,
+      },
+    },
+  },
+  responses: {
+    "201": {
+      description: "Comment created successfully — returns the comment",
+      content: {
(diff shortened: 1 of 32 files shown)


apps/backend/src/api/schema/primitives/buildReview.ts
```diff
@@ -0,0 +1,48 @@
...
+  });
+
+export const BuildReviewSchema = z
+  .object({
+    id: z.string(),
+    buildId: z.string(),
+    state: ReviewStateSchema,
+    userId: z.string().nullable().meta({
+      description: "ID of the user who submitted the review.",
+    }),
```
gregberge: Same we should replace by `user` and put the user.

apps/backend/src/api/schema/primitives/buildReview.ts
```diff
@@ -0,0 +1,48 @@
...
+    dismissedAt: z.string().nullable().meta({
+      description: "Date the review was dismissed, null if not dismissed.",
+    }),
+    dismissedById: z.string().nullable().meta({
+      description: "ID of the user who dismissed the review, if any.",
+    }),
```
gregberge: Same we should replace by `dismissedBy` and put the user.

apps/backend/src/api/handlers/createComment.ts
```diff
@@ -0,0 +1,214 @@
...
+import {
+  forbidden,
+  invalidParameters,
+  notFound,
+  serverError,
+  unauthorized,
+} from "../schema/util/error";
+import { CreateAPIHandler } from "../util";
+
+const AnchorInputSchema = z
```
gregberge: Why are we doing this and not using the same schema as the one used in comment output? in `schema/primitives`?

apps/backend/src/api/schema/primitives/comment.ts
```diff
@@ -0,0 +1,206 @@
...
+
+const CommentReactionGroupSchema = z
+  .object({
+    emoji: z.string().meta({ description: "The emoji used for the reaction." }),
+    count: z
+      .number()
+      .meta({ description: "Number of users who reacted with this emoji." }),
+    userIds: z
+      .array(z.string())
+      .meta({ description: "IDs of the users who reacted with this emoji." }),
```
gregberge: We should probably list all users?

apps/backend/src/comment/body.ts
```diff
@@ -0,0 +1,47 @@
...
+ * Markdown — use the raw-JSON form to mention users.
+ *
+ * Length/empty/sanitize limits are enforced downstream by `createBuildComment`,
+ * `updateBuildComment` and `createBuildReview`; this only produces the document.
+ */
+export function resolveCommentBody(input: string | object): JSONContent {
+  if (typeof input === "string") {
+    // marked is synchronous unless async extensions are registered; force the
+    // sync return so the type is a plain string rather than a promise.
+    const html = marked.parse(input, { async: false });
```
gregberge: Not sure it's a good idea to use it sync, switch to async.

apps/backend/src/api/schema/primitives/buildReview.ts
gregberge: Done in 303bfeb5 — replaced `userId` with an embedded `user` object, resolved through a new shared `UserSchema` primitive (batched to avoid N+1).

apps/backend/src/api/schema/primitives/buildReview.ts
gregberge: Done in 303bfeb5 — now exposes a `dismissedBy` user object (same `UserSchema`).

apps/backend/src/api/handlers/createComment.ts
gregberge: Good call — done in 72be334d. Dropped the bespoke `{ point?, lines? }` input and reuse the `CommentAnchorSchema` exported from `schema/primitives/comment`, so input and output share the same shape. Bounds and the inverted-range rule are still validated against the model schema and surfaced as a 400.

apps/backend/src/api/schema/primitives/comment.ts
gregberge: Done in 1ed4f9de — reactions now return `users` (full user objects) instead of `userIds`, reusing the same batched account lookup as the author.

apps/backend/src/comment/body.ts
gregberge: Done in 7e4b5bc2 — `resolveCommentBody` is now async and awaits `marked.parse`; the three API callers await it.

Issue #1631, opened 2025-05-21
Flaky management flow


gregberge:
Allow to mark a test as flaky, to resolve its flakiness and automatically detect regressions of a flaky test and adverts users about it.

gregberge:
Still planed but we use Linear to track this.

PR #2288, opened 2026-06-23
feat(account): add account-wide flaky Tests page

ton03:
## What

Adds an **account-wide Tests page** that aggregates the active tests across *all* of an account's visible projects into a single flakiness-sorted list, surfaced as a new **Tests** tab on the account (next to Analytics).

Today flakiness is only visible one project at a time (the per-project Tests page). On accounts with many projects there's no way to answer "which are my flakiest tests overall?" without opening each project in turn. This adds that org-wide view.

## Screenshots

| Account-wide Tests page (`/:accountSlug/~/tests`) | New "Tests" tab |
| --- | --- |
| _flaky-sorted list across all projects, with a Project column_ | _Projects · Analytics · Tests · Settings_ |

## How

The per-project Tests resolver already sorted by a read-time flakiness score whose ordering subquery is correlated only on `tests.id` + the period bounds — i.e. it never depended on how many projects the candidate set spanned. So the core change is small:

- **Extracted `queryActiveTests({ projectIds, period, filters, after, first })`** into `graphql/services/test.ts`, holding the active-test detection (latest reference build per build-name → non-orphan compare screenshots) and the flakiness `orderByRaw`. `Project.tests` now calls it with `[project.id]`; the SQL is unchanged (verbatim move), so per-project behaviour is identical.
- **`Account.tests`** field + resolver: resolves the viewer's visible project ids, then calls `queryActiveTests` with `whereIn(projectIds)`. Empty visibility short-circuits to an empty connection.
- **`getVisibleProjectIds({ account, user })`** in `graphql/services/project.ts` applies the same visibility rules as `Account.projects` (staff see all; user account → owner only; team → owner/member see all, contributor sees their projects or those with a default user level).
- **`Test.project`** field added so the new list can show which project each row belongs to.
- The `tests` field is declared on the `Account` interface and mirrored on the `Team`/`User` implementing types (GraphQL requires implementing types to declare interface fields); the resolver is shared via `commonAccountResolvers`.
- Frontend: new `pages/Account/Tests.tsx` (reuses the per-project list primitives + adds a Project column linking back to each project), a `~/tests` route, and the nav tab.

## A note on performance

The flakiness score is computed per candidate test at read time (the ordering subquery scans `test_stats_builds` / `test_stats_fingerprints` over the period). Per-project that's bounded by one project's test count; account-wide it's bounded by every active test across all of the account's projects, so on large accounts this is a heavier single query. It's still one query (no N+1), and the `period` filter bounds the stats scan, but the long-term fix for flakiness-at-scale is a materialized rollup refreshed on build finalize — cf. #1631. Happy to follow up with that separately if you'd like it before this lands.

## Testing

- New `graphql/tests/accountTests.e2e.test.ts`: seeds two projects with differing flakiness (driven via `upsertTestStats`) and asserts the list aggregates account-wide, sorts flaky-first across project boundaries, resolves the owning project per row, and returns an empty connection when the account has no visible projects.
- `pnpm run lint`, `pnpm run check-types`, and `pnpm run codegen` are clean.
- Verified locally end-to-end against a seeded Postgres (the screenshots above are the real page rendering).

---

### DIFF
--- a/apps/backend/src/graphql/definitions/Account.ts
+++ b/apps/backend/src/graphql/definitions/Account.ts
@@ -37,6 +37,8 @@ import {
 import type { Context } from "../context";
 import { getAdminAccount } from "../services/account";
 import { getAccountAvatar } from "../services/avatar";
+import { getVisibleProjectIds } from "../services/project";
+import { queryActiveTests } from "../services/test";
 import { badUserInput, unauthenticated } from "../util";
 import { paginateResult } from "./PageInfo";
 
@@ -137,6 +139,13 @@ export const typeDefs = gql`
     hasForcedPlan: Boolean!
     permissions: [AccountPermission!]!
     projects(after: Int = 0, first: Int = 30): ProjectConnection!
+    "List all tests across the account's visible projects, sorted by flakiness."
+    tests(
+      after: Int = 0
+      first: Int = 30
+      period: MetricsPeriod!
+      filters: TestsFilterInput
+    ): TestConnection!
     avatar: AccountAvatar!
     gitlabAccessToken: String
     gitlabBaseUrl: String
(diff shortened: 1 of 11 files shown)


gregberge:
Thanks for the contribution, I opened another PR to continue the work.

PR #2290, opened 2026-06-25
feat: add project ignore configuration

gregberge:
## Goal

Let users control the ignore feature per project. Some users want to disable it completely.

Fix ARG-414

### Data model
- Replace the project `autoIgnore` column with a flexible `ignoreConfig` jsonb column. Only values that differ from the default are stored. The default is **ignore feature enabled + auto-ignore enabled at 3 occurrences**.
  - absent / `null` → default
  - `{ "enabled": false }` → ignore feature disabled
  - `{ "autoIgnore": false }` → auto-ignore disabled
  - `{ "autoIgnore": { "changes": N } }` → custom threshold
- Migration transforms existing data (`null` → auto-ignore was off, `{changes:3}` → default, `{changes:N}` → custom). The legacy `autoIgnore` column is intentionally **kept** and will be dropped in a follow-up PR.

### Backend behavior
- `ignoreChange` / `unignoreChange` mutations are rejected when the feature is disabled.
- Auto-ignore is disabled when the feature is disabled.
- Already-ignored diffs **stay ignored** even when the feature is disabled (no retroactive effect on previous builds).
- Expose `ignoreConfig` in the GraphQL `Project` type + `updateProject` input (normalized to minimal storage).

### Frontend
- Restructure the project settings (Flaky detection): a global "Ignore changes" enable toggle on top, with the auto-ignore card hidden when disabled (same pattern as Deployments).
- Hide the ignore button in the build/test views when the feature is disabled, via a small `ProjectIgnoreEnabledProvider` context.

## Tests
- New unit tests for the `resolveIgnoreConfig` / `normalizeIgnoreConfig` helpers (round-trip + default behavior).
- Existing e2e tests pass (computeScreenshotDiff, getBuildDiffs, approval, bucket-merge). Typecheck + ESLint clean on both apps.


jsfez (approved):

apps/backend/src/graphql/definitions/Project.ts
```diff
@@ -497,11 +512,9 @@ const importGitlabProject = async (props: {
 function createProject(
```
gregberge: I think we can remove this func, it's useless now.

gregberge: Good catch — removed the wrapper and inlined `Project.query().insertAndFetch(...)` at the three call sites in fdc0b6d3.

PR #2328, opened 2026-07-02
fix(build): render browser icon eagerly to stop flaky screenshots

jsfez:
## Problem

The Chromium browser icon in the build metadata panel produced flaky visual test diffs.

`BrowserRow` was the only consumer wrapping `BrowserIcon` in `lazy()` + `Suspense`. The Suspense fallback is a fully-rendered `GlobeIcon` — a "complete" element with no loading signal — so Argos's stabilization had no reason to wait, and would screenshot **either** the globe fallback **or** the real icon depending on when the lazy chunk resolved. Hence the flake.

The other two consumers ([`ContextSection`](../blob/main/apps/frontend/src/pages/Build/BuildOverview/ContextSection.tsx) and [`FilterIcon`](../blob/main/apps/frontend/src/pages/Build/metadata/filters/FilterIcon.tsx)) import `BrowserIcon` directly and don't flake.

## Fix

Import `BrowserIcon` directly in `BrowserRow` instead of lazy-loading it. The component is trivial (a handful of small SVG imports), so lazy-loading offered no real benefit and only introduced the async race.

The unknown-browser `GlobeIcon` fallback inside `BrowserIcon` itself is unchanged — only the transient loading fallback is gone.

## Notes

`aria-busy` on the fallback would also have fixed this, but removing the pointless lazy import addresses the root cause rather than working around screenshot timing.

### DIFF
--- a/apps/frontend/src/pages/Build/sidebar/metadata/BrowserRow.tsx
+++ b/apps/frontend/src/pages/Build/sidebar/metadata/BrowserRow.tsx
@@ -30,7 +30,7 @@ const LazyBrowserIcon = lazy(() =>
 function BrowserIcon(props: { browser: MetadataBrowser; className?: string }) {
   const { browser, ...rest } = props;
   return (
-    <Suspense fallback={<GlobeIcon {...rest} />}>
+    <Suspense fallback={<GlobeIcon aria-busy {...rest} />}>
       <LazyBrowserIcon browser={browser} {...rest} />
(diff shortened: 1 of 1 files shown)


gregberge (changes requested):
It's not the good fix. We will not impact perf for visual testing, the good fix is to put a aria-busy on the fallback.

gregberge (approved):

PR #2336, opened 2026-07-05
feat(frontend): rework rejection note dialog (ARG-444)

gregberge:
## Summary

Reworks the **"Add a note about this rejection"** dialog (shown when rejecting a snapshot during build review) per ARG-444.

- **Normal button in the footer, not in the field** — replaces the editor's baked-in send button with a normal **"Add to review"** button in the dialog footer, next to **Skip**. ⌘/Ctrl+Enter still submits.
- **Snapshot shown on top of the field** — a thumbnail + name of the snapshot the note attaches to, so the reviewer sees what they're commenting on (mirrors the sidebar's snapshot reference).
- **Post-submit flow** — once the comment is added: the dialog closes, the right sidebar opens on the **Review** panel (so the new comment is visible), and the viewer advances to the next snapshot to keep reviewing.

Empty submit still surfaces the "Comment required" toast, and content is preserved on error. The provider/context/hook API is unchanged, so `TrackButtons`/`BuildPage` need no changes.

## Notes

- **Skip** closes without advancing (only submit advances), matching the issue wording.

## Test plan

- [ ] Reject a snapshot → dialog shows the snapshot thumbnail + name above the field, and an "Add to review" button in the footer
- [ ] Type a note and submit → comment is added, right sidebar opens on the Review panel showing the comment, viewer advances to the next snapshot
- [ ] Submit while empty → "Comment required" toast, no comment created
- [ ] Skip → dialog closes, snapshot stays rejected

### DIFF
--- a/apps/frontend/src/pages/Build/RejectCommentDialog.tsx
+++ b/apps/frontend/src/pages/Build/RejectCommentDialog.tsx
@@ -4,9 +4,10 @@ import {
   useCallback,
   useId,
   useMemo,
+  useRef,
   useState,
 } from "react";
-import { useMutation } from "@apollo/client/react";
+import { useApolloClient } from "@apollo/client/react";
 import { invariant } from "@argos/util/invariant";
 import { toast } from "sonner";
 
@@ -23,7 +24,7 @@ import {
 } from "@/ui/Dialog";
 import { Editor, type EditorValue } from "@/ui/Editor/Editor";
 import { hasEditorContent } from "@/ui/Editor/util";
-import { Modal } from "@/ui/Modal";
+import { Modal, ModalActionContext } from "@/ui/Modal";
 import { getMentionUser } from "@/ui/UserCard";
 import { getErrorMessage } from "@/util/error";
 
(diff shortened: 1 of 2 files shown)


apps/frontend/src/pages/Build/RejectCommentDialog.tsx
```diff
@@ -127,32 +139,64 @@ function RejectCommentDialog(props: {
...
+    null;
   const mentions = useMemo(
     () => build.members.map(getMentionUser),
     [build.members],
   );
+  const [value, setValue] = useState<EditorValue>(null);
+  const [isPending, setIsPending] = useState(false);
+  const emptyToastId = useId();
+  const isEmpty = !hasEditorContent(value);
   const [addBuildComment] = useMutation(AddBuildCommentMutation);
```
gregberge: Do not use this, you can use directly the Apollo Client in this case. Having a `useMutation` is useless if you don't want to track the progression directly. It re-renders the component for nothing while in progress.

Mention this in AGENTS.md also.

PR #2352, opened 2026-07-10
feat(backend): add API endpoint to create a project (ARG-451)

gregberge:
## What

Adds `POST /v2/projects` to the OpenAPI (v2) API to create a project, mirroring the existing GraphQL `createProject` mutation.

- **Auth:** personal access token that must be **scoped to the target account** and belong to one of its **administrators**.
- **Body:** `{ name, accountSlug }` → **`201`** with the serialized project.
- **Errors:** `400` (invalid/reserved/duplicate name), `401` (bad token / account not in token scope), `403` (not an admin), `500`.

## How

Two pieces of shared logic were extracted so the GraphQL and REST surfaces stay in sync:

**Shared project-name schema** — `ProjectNameSchema` (Zod: trim, 1–100 chars, `^[a-zA-Z0-9_\-.]+$`) is the single source of truth in `database/util/project-name.ts`. It feeds:
- the `Project` model's JSON schema via `z.toJSONSchema(...)` (same pattern already used for `timestampsSchema`/`UserLevelJsonSchema`), and
- the REST operation's request body (so the OpenAPI docs now advertise the pattern/length too).

**Shared `createProject` service** — `database/services/project.ts` now owns the creation logic (admin-permission check, name validation, reserved/uniqueness check, insert, Discord notification) and throws `boom`, matching the existing shared-service convention (`createBuildReview`). Consumed by:
- the **REST handler** (thin: resolves the account from token scope, then calls the service), and
- the **GraphQL resolver** (keeps `unauthenticated()` + account resolution, then translates the service's `boom` back into the existing GraphQL error contract — `403 → FORBIDDEN`, `400 → BAD_USER_INPUT{field:"name"}` — so nothing changes for existing clients).

A new `getAccountForAuth` helper resolves a PAT-scoped account (an account outside the token's scope is indistinguishable from one that doesn't exist — both `401` — so existence is never disclosed).

## Scope

This covers the **API** half of ARG-451. The CLI half is not included here.

## Testing

- `database/services/project.e2e.test.ts` — service unit coverage (success, trim, non-admin `403`, invalid/reserved/duplicate name).
- `api/handlers/createProject.e2e.test.ts` — REST integration (auth, token scope, admin, name rules, serialization).
- Existing GraphQL `createProject` e2e tests pass unchanged, confirming the error contract is preserved.
- Lint, typecheck, and prettier are green.


packages/schemas/src/project.ts
gregberge, line 1: Move this to https://github.com/argos-ci/argos/tree/main/packages/schemas

apps/backend/src/graphql/definitions/Project.ts
```diff
@@ -912,27 +892,26 @@ export const resolvers: IResolvers = {
...
+      } catch (error) {
+        // The shared service speaks HTTP status codes; map them onto this
+        // API's GraphQL error contract.
+        if (error instanceof HTTPError) {
+          if (error.statusCode === 403) {
+            throw forbidden(error.message);
+          }
+          if (error.statusCode === 400) {
+            throw badUserInput(error.message, { field: "name" });
+          }
```
gregberge: Use an error code for this, a 400 does not mean it concerns the name.

PR #2357, opened 2026-07-11
feat: expose account analytics via the REST API (ARG-461)

gregberge:
## What

Exposes account analytics through the REST API and unifies the metrics layer behind a single entry point shared by GraphQL and REST.

## Changes

- **New REST endpoint** `GET /accounts/{accountSlug}/analytics` ([getAccountAnalytics.ts](apps/backend/src/api/handlers/getAccountAnalytics.ts)) returning build and screenshot metrics. Authenticated with a personal access token scoped to the account, documented under a new **Analytics** OpenAPI tag.
- **Shared `getAccountMetrics()` entry point** ([account.ts](apps/backend/src/metrics/account.ts)) used by both GraphQL and REST. It resolves project names to IDs, validates the date range (throwing `InvalidAccountMetricsInputError`, translated to `400` / `BAD_USER_INPUT` by each caller), and delegates to the existing per-metric aggregations.
- **Project-name filtering** — metrics can now be filtered by `projectNames` in addition to internal `projectIds`. The GraphQL `metrics` input gains a `projectNames` field, and the frontend ([Analytics.tsx](apps/frontend/src/pages/Account/Analytics.tsx)) filters by name directly instead of pre-resolving names to IDs with an extra query.
- **GraphQL resolver** ([Account.ts](apps/backend/src/graphql/definitions/Account.ts)) now delegates to the shared entry point, dropping its duplicated validation logic.

## Tests

- REST handler e2e tests: name filtering, no-match, invalid date range (`400`), unscoped token (`401`), project-token rejection.
- GraphQL e2e tests: filtering by name, filtering by ID (regression), invalid date range as `BAD_USER_INPUT`.
- Metrics unit tests for `getAccountMetrics` (name filtering, no-match) and empty-`projectIds` behavior.

### DIFF
--- a/apps/backend/src/metrics/account.ts
+++ b/apps/backend/src/metrics/account.ts
@@ -7,7 +7,6 @@ type AccountMetricsGroupBy = "month" | "week" | "day";
 
 type AccountMetricsFilter = {
   accountId: string;
-  projectIds?: string[] | null | undefined;
   projectNames?: string[] | null | undefined;
 };
 
@@ -43,14 +42,7 @@ function hasProjectFilter(input: AccountMetricsAggregationInput) {
 }
 
 async function resolveProjectIds(input: AccountMetricsFilter) {
-  if (input.projectNames === null || input.projectNames === undefined) {
-    return {
-      projectIds: input.projectIds,
-      projectFilterApplied: Boolean(input.projectIds?.length),
-    };
-  }
-
-  if (input.projectNames.length === 0) {
+  if (!input.projectNames || input.projectNames.length === 0) {
     return { projectIds: undefined, projectFilterApplied: false };
   }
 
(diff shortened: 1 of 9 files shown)


apps/backend/src/metrics/account.ts
```diff
@@ -1,18 +1,114 @@
...
+type AccountMetricsGroupBy = "month" | "week" | "day";
 
-/**
- * Get the number of screenshots by projects over time.
- */
-export async function getAccountScreenshotMetrics(input: {
+type AccountMetricsFilter = {
+  accountId: string;
+  projectIds?: string[] | null | undefined;
+  projectNames?: string[] | null | undefined;
```
gregberge: why are we accepting both here?

gregberge: Good catch — no real reason to keep both. After this PR the frontend and the REST API both filter by name, so nothing sends `projectIds` anymore; the only thing exercising it was a regression test I'd added.

Dropped `projectIds` from the filter and the GraphQL `AccountMetricsInput`, standardizing on `projectNames` as the single public key. Internal project IDs are still resolved from names in `resolveProjectIds` and used for the SQL aggregation. Done in 1354f019.

PR #2373, opened 2026-07-21
fix(billing): convert trial to paid plan when usage exceeds included screenshots

jsfez:
## Problem

Stripe never bills metered usage consumed during a trial period. `checkIsOutOfCapacity` treated "trialing + payment method filled" as never out of capacity, assuming overage would be billed — it never was. A customer consumed **181,384 overage screenshots** during their trial and was invoiced **$0** for them ("Trial period for Extra screenshot" line at $0.00).

Worse, the incentive was inverted: a trial **without** a card was capped at the included screenshots, while adding a card unlocked unlimited free usage.

## Fix

- `checkIsOutOfCapacity` now reports over-capacity trials regardless of payment method.
- On build creation, `endTrialToUnlockUsage`:
  - **card on file** → ends the Stripe trial (`trial_end: "now"`) under a redis lock. A fresh billing period opens (`startDate` synced from `current_period_start`), so screenshots consumed during the trial stay unbilled and the team restarts with its full included quota.
  - **no card** → build rejected with the existing 402.
- **Idempotent conversion**: Stripe is retrieved first inside the lock; a trial already ended (timed-out previous attempt, natural expiry not yet webhooked) is synced instead of ended again — re-setting `trial_end` could start a new trial.
- **Owners are emailed** (`trial_ended`, transactional `account` category, not opt-out-able) announcing the plan start and upcoming charge. Discord + email notifications run **outside the lock**, so the critical section only holds Stripe calls.
- **Error policy**: deterministic errors (unmapped Stripe product, malformed price metadata, deleted row) surface loudly; anything else (Stripe/Redis outage) fails open with Sentry context so build creation never depends on Stripe availability.
- New `queue-pending-notification-workflows` bin (same pattern as the three existing `queue-pending-*` bins, plus a stuck-`pending` branch) recovers notification workflows persisted but never pushed to the queue.

## Testing

- 12 e2e tests on `endTrialToUnlockUsage` (conversion, idempotent recovery, canceled-mid-race, no-card, non-Stripe, fail-open on outage, deterministic error surfacing, notification failure tolerance, no lock churn).
- 2 e2e tests on the `createBuild` wiring (conversion → build created + email; no card → 402), mutation-checked: inverting the branch fails exactly these two.
- 5 e2e tests on `checkIsOutOfCapacity` (previously untested).
- Period reset asserted (`startDate` lands on the conversion timestamp) — a regression re-counting trial screenshots into the first billed period now fails the suite.
- Render test for all 16 notification handlers.

## Known follow-ups (deliberately out of scope)

- Spend-limit check still evaluated pre-conversion in the same request (accepted edge case: one spurious 402 right after conversion, self-heals).
- `notification_workflows` has no secondary index; the new bin's query seq-scans (needs a small migration).
- Natural trial expiry (webhook path) still sends no email — needs a dedup mechanism/migration to hang it on the status transition.
- `endTrialToUnlockUsage` returns a bare boolean; a discriminated outcome would make outage fail-opens observable at the call site.


apps/backend/src/notification/bin/queue-pending-notification-workflows.ts
gregberge, line 1: Why this file? Let's remove it.

apps/backend/src/notification/handlers/handlers.test.ts
gregberge, line 1: This file it out of context also.

apps/backend/src/notification/handlers/trial_ended.tsx
```diff
@@ -0,0 +1,73 @@
...
+  Signature,
+} from "../../email/components";
+import { defineNotificationHandler } from "../workflow-types";
+
+const baseUrl = config.get("server.url");
+
+export const handler = defineNotificationHandler({
+  type: "trial_ended",
+  // Announcing a charge is transactional, it must not be opt-out-able.
+  category: "account",
```
gregberge: Nop we should not be able to disable notification for this.

gregberge (approved):

apps/backend/src/notification/bin/queue-pending-notification-workflows.ts
jsfez: Removed — you're right, recovering stranded notification workflows is a general concern, not this PR's. Worth noting it leaves one gap open: `sendNotification` commits the workflow row before pushing to RabbitMQ, so a failure in that window drops the email silently (only Sentry sees it). Happy to open a separate issue if you want it tracked.

apps/backend/src/notification/handlers/handlers.test.ts
jsfez: Removed. It was rendering all 16 handlers from their `previewData`, which is broader than this PR — agreed it belongs elsewhere. The `trial_ended` template itself stays covered indirectly (`trial.e2e.test.ts` asserts the notification is dispatched with the right payload), just not its rendered output.

apps/backend/src/notification/handlers/trial_ended.tsx
jsfez: Agreed on the requirement — and that's already the behaviour: `account` is a **non-configurable** category ([categories.ts](https://github.com/argos-ci/argos/blob/main/apps/backend/src/notification/categories.ts#L28-L32) sets `configurable: false`), which is exactly why I picked it over `billing`.

Concretely it means:
- `workflow-job.ts` only builds the opted-out set for configurable categories, so no recipient is ever filtered out for this one.
- `message-job.ts` resolves `preferencesUrl` to `null`, so the email carries no unsubscribe link.
- `getConfigurableNotificationCategories()` excludes it, so it never shows up in the preferences UI.

If it had been `billing` (which *is* configurable) a customer could have muted a charge notice — that's the trap I was avoiding. I've reworded the comment to name the mechanism instead of just the intent, since it clearly wasn't obvious from the code.

PR #2374, opened 2026-07-22
feat(discord): route project creations to a dedicated growth channel

jsfez:
## Problem

Project creation notifications are currently sent to the Discord channel used for human-to-human communication. Muting that channel would also hide legitimate conversations, while the notification volume continues to grow with the number of projects. As a result, important but infrequent notifications—such as subscription changes and trial extensions—are increasingly buried.

## Solution

The single webhook is being split into two named channels:

- **`head`** — rare events that warrant an interruption, such as subscription changes and trial extensions.
- **`growth`** — an activity feed intended to be muted and checked on demand, such as project creation events.

The `channel` argument is now required instead of being optional with `head` as the default. If a future caller forgets to provide it, the request will fail rather than silently sending notifications to the channel that cannot be muted—the exact problem this change is intended to solve.

`DISCORD_WEBHOOK_URL` has been renamed to `DISCORD_HEAD_WEBHOOK_URL` to follow the existing `DISCORD_<CHANNEL>_WEBHOOK_URL` convention used by `DISCORD_RELEASE_WEBHOOK_URL` in the release workflow.

CleanShot 2026-07-22 at 13 41 44@2x

**Note:** Once the deployment has been confirmed, the now-orphaned `/production/discord_webhook_url` parameter can be safely deleted.


gregberge:
We will not create a channel dedicated to notifications.

PR #2376, opened 2026-07-22
feat(discord): notify when a project is transferred to a team

jsfez:
Notifies when a project moves from a personal account to a team — the moment a side project becomes something a team relies on. Every other direction stays silent.

Both accounts are read before the patch: `patchAndFetch` mutates the instance in place, overwriting the source account id and the previous name. A test covers that ordering.

### DIFF
--- a/apps/backend/src/graphql/definitions/Project.ts
+++ b/apps/backend/src/graphql/definitions/Project.ts
@@ -1114,18 +1114,20 @@ export const resolvers: IResolvers = {
         accountId: targetAccountId,
       });
 
-      // Load both accounts before patching: `patchAndFetch` mutates the
-      // instance in place, overwriting the source account id and the old name.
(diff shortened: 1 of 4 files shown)


apps/backend/src/graphql/definitions/Project.ts
```diff
@@ -1112,10 +1113,31 @@ export const resolvers: IResolvers = {
...
+      const [previousAccount, targetAccount] = await Promise.all([
+        ctx.loaders.Account.load(project.accountId),
+        ctx.loaders.Account.load(targetAccountId),
+      ]);
+      const previousName = project.name;
+
+      const transferredProject = await project.$query().patchAndFetch({
         accountId: targetAccountId,
         name: args.input.name,
       });
```
gregberge: This should be loaded in parallel of the `patchAndFetch` to save some time.

gregberge (approved):

PR #2377, opened 2026-07-22
feat(staff): add a trial pipeline dashboard

jsfez:
Adds a `/staff` section with tabs holding the existing directory and a new **Trial pipeline**: teams created over a rolling window, how far each got, their billing state and usage. A mailto button drafts the onboarding email from the row's own state (orphan builds vs welcome), and a switch records that it went out.

`teams/all` redirects to `/staff/teams`. Owner emails and `User.staff` are gated on the field itself, not just the parent query.

CleanShot 2026-07-22 at 22 58 46@2x


apps/backend/src/graphql/definitions/Team.ts
```diff
@@ -125,6 +132,16 @@ export const typeDefs = gql`
...
+    projectsCount: Int!
+    buildsCount: Int!
+    "Screenshots uploaded since the team was created"
+    screenshotsCount: Int!
+    "When the team got its first build compared to a baseline"
+    firstComparisonAt: DateTime
+    "Owners of the team, with their email address. Staff only."
+    staffOwners: [StaffTeamOwner!]!
+    "When a staff member reached out to the team, null if never. Staff only."
+    staffContactedAt: DateTime
```
gregberge: Put all these fields inside a "staff" field accessible only by staff. For example here you leaked the buildsCount publicly. It's easy to make a mistake.

apps/backend/src/graphql/definitions/Team.ts
```diff
@@ -152,6 +169,18 @@ export const typeDefs = gql`
     githubLightInstallation: GithubInstallation
   }
 
+  "An owner of a team, as needed to write to them. Staff only."
+  type StaffTeamOwner {
```
gregberge: Isolate all the staff stuff in a dedicated file in the GraphQL folder.

apps/backend/src/graphql/definitions/Team.ts
```diff
@@ -346,13 +375,17 @@ export const typeDefs = gql`
     teamInvite(secret: String!): Team
     "List all teams (staff only)"
     staffTeams: [Team!]!
+    "List teams created within the last \`days\` days, newest first (staff only)"
+    staffTrialPipeline(days: Int! = 30): [Team!]!
```
gregberge: Same for queries all the staff in a file.

apps/backend/db/migrations/20260722120000_account-staff-contacted-at.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * @param {import('knex').Knex} knex
+ */
+export const up = async (knex) => {
+  await knex.schema.alterTable("accounts", (table) => {
+    table.dateTime("staffContactedAt");
```
gregberge: I am a bit worried to have CRM related data in our database. Should we at least use another table to store it? I think what we need here is more a row that represents a contact than just the date of the contact?

gregberge (dismissed):
I am wondering it what you are doing here should not be built in a CRM instead.

apps/backend/db/migrations/20260722120000_account-staff-contacted-at.js
jsfez: You are right. I wanted to reduce the surface of the migration and avoid creating a whole table. But it will be cleaner to a one.

jsfez:
> I am wondering it what you are doing here should not be built in a CRM instead.

I don’t see why storing this data here would be an issue. In a CRM, I would have to create a new entry for each team, which seems like a lot of work just to check a box.

That said, for enterprise negotiations or more complex workflows, the CRM is the best way to manage the lead.

PR #2381, opened 2026-07-23
fix(billing): say why an add-on cannot be enabled

jsfez:
Stacked on #2379.

Add-ons are billed as extra products on a Stripe subscription. A forced plan has none, and a GitHub Marketplace one cannot carry them — both were offered an Enable button that the API then refused, or a message naming the wrong reason. Neither was an access hole: the API already refused both.

The rule moves into `getAddOnBlockedReason`, shared by the three add-on screens, with the mutations returning the same wording. Its order matters: a plan that already includes the feature answers before billing is asked about, so an account that owns it without a Stripe subscription keeps it.

### DIFF
--- a/apps/frontend/src/util/subscription.ts
+++ b/apps/frontend/src/util/subscription.ts
@@ -1,0 +1,48 @@
+import { checkIsActiveSubscriptionStatus } from "@argos/schemas/subscription-status";
+
+import {
+  AccountSubscriptionProvider,
+  AccountSubscriptionStatus,
+} from "@/gql/graphql";
+
+/**
+ * Why a paid add-on cannot be turned on, or null when it can.
+ *
+ * The order of these checks is the point. A plan that already carries the
+ * feature answers before anything about billing is asked — otherwise accounts
+ * that legitimately have it, but no Stripe subscription to bill an add-on to,
+ * would be refused the feature they already own.
+ *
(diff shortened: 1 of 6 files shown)


apps/frontend/src/util/subscription.ts
```diff
@@ -0,0 +1,48 @@
...
+  // forced plan has none, and a GitHub Marketplace subscription cannot carry
+  // them — both are healthy plans that simply have nothing to bill against.
+  if (provider !== AccountSubscriptionProvider.Stripe) {
+    return `Your plan does not allow enabling ${featureName}, please contact us.`;
+  }
+
+  if (!usageBased) {
+    return "This feature is not available on your current plan, please contact us.";
+  }
+
```
gregberge, line 61: Check also yearly.

gregberge (approved):

apps/frontend/src/util/subscription.ts
jsfez: Done in 1198725. Add-on products are priced monthly and Stripe refuses an item whose interval differs from the rest of the subscription, so a yearly plan cannot carry one either. `Plan.interval` is now exposed in GraphQL, `getAddOnBlockedReason` returns "<feature> cannot be added to a yearly plan, please contact us.", and both mutations throw the same wording.

PR #2393, opened 2026-07-24
feat(staff): reach out to lost teams and count carded trials as converted

jsfez:
The staff pages move into a `Staff` folder, next to their layout.

A team whose trial ran out or whose subscription was canceled now gets an email asking why it stopped rather than a welcome, and both tables link to their Stripe customer. A trial with a payment method counts as converted right away — the rate denominator gets a tooltip since that rule is written nowhere else.


apps/frontend/src/pages/Staff/Trials.email.ts
```diff
@@ -25,23 +25,46 @@ function getGreeting(owners: Owner[]): string {
...
+        "I noticed that you stopped testing Argos, and I'd love to understand why.",
+        "",
+        "Were you testing a specific use case, or did you run into any blockers during setup?",
+        "",
+        "Any feedback would be greatly appreciated.",
+        "",
+        "Have a great day,",
+      ].join("\n"),
+    };
+  }
```
gregberge, line 61: You can't send this kind of email on a `contact@argos-ci.com`, it's not enough personal and will not be understood.

gregberge (changes requested):

apps/frontend/src/pages/Staff/Trials.email.ts
jsfez: This email won't be send from Argos. It's a mailto link with a prefilled content

gregberge: OK sorry.

gregberge (approved):

PR #2394, opened 2026-07-24
feat(msteams): send build notifications to Microsoft Teams channels

jsfez:
Adds a Microsoft Teams integration built on Power Automate Workflows incoming webhooks: an account registers one webhook URL per channel, and a new `sendMsTeamsMessage` automation action posts an Adaptive Card mirroring the Slack message.

Webhook hosts are allowlisted — the URL is user-supplied and posted to by our servers. Validated end to end against a real M365 tenant: the URL Teams hands out today is on `*.environment.api.powerplatform.com`, not `logic.azure.com` as the docs suggest.

Adding a second automation action also forced a rework of the `AutomationActionRun` JSON Schema — Objection strips `required` recursively on partial updates, so a two-branch `oneOf` matched every `patch()` vacuously and broke the job pipeline.

No link unfurling: a webhook is outbound-only, so Argos never sees messages posted in Teams. That would require a full Teams app.

https://claude.ai/code/session_014YTG24WUnBtJjxTPTMWroH


gregberge (commented):
It looks good but I have to test it.

apps/backend/src/automation/actions/index.ts
```diff
@@ -33,3 +82,59 @@ export function getAutomationAction<T extends AutomationActionType>(
...
+    case "sendMsTeamsMessage":
+      await sendMsTeamsMessage.automationAction.test({
+        payload: action.actionPayload,
+        message,
+      });
+      return;
+    default:
+      assertNever(action);
+  }
+}
```
gregberge, line 128: This code could probably be simplified, we could just get the automationAction from the `action.action` so the code will be factorized.

gregberge: Done in e5a94fd8 — both dispatchers now look the handler up with `getAutomationAction(action.action)`, same as `triggerAutomation` does, so adding an action no longer means editing two switches.

One thing to note: the lookup by name is what forced the switch in the first place. TypeScript widens `process`/`test` to a union of signatures and then only accepts a payload satisfying *every* action at once, and it cannot correlate `action` with `actionPayload` across a union without narrowing. The payload therefore goes through a single documented `asHandlerPayload` assertion — `AutomationActionSchema` has already validated it against that exact action, so the pairing is guaranteed at runtime.

Side effect: exhaustiveness is no longer checked by the compiler (`assertNever` is gone); a name missing from `AUTOMATION_ACTIONS` now throws at runtime instead. That is already how `triggerAutomation` behaves, and a new action has to be added to `AutomationActionSchema` to be a valid type at all.

PR #2398, opened 2026-07-26
feat(onboarding): welcome new signups with a first-run screen

gregberge:
New accounts landed on the dashboard with nothing asked of them. They now land on `/~/welcome`.

CleanShot 2026-07-26 at 23 11 56@2x

## What it asks

**Where did you find Argos?** — search engine / AI assistant / social media / GitHub / word of mouth / somewhere else (+ free text). Answers land on `users.signupSource`.

**Domain auto-join** — when a team was just created and the user has a verified address on a domain not just anyone can sign up to, the page offers to let that domain join the team. Reuses the existing `team_domains` feature. `/teams/new` grows the same checkbox for teams created later.

Both questions are skippable.

## Where the hop is inserted

```
hobby:   signup → oauth → /~/welcome ─────────────────→ /:slug
team:    signup → oauth → /teams/new (auto) → Stripe → /~/welcome → /:team
joining: signup → oauth → /~/welcome ─────────────────→ /teams
```

Auth wraps the post-signup destination, **except** when that destination auto-creates a team — there `createTeam` inserts the hop itself, once the team exists, because the domain question has nothing to attach to before then.

`signupSourceAskedAt` is stored apart from the source so a skip is distinguishable from never having asked, and it is what stops the page reappearing on the user's next team.

`/~/welcome` sits under `~/` so it can never collide with an account slug — no new reserved slug.

## Consumer email domains

Eligibility requires a domain that isn't a shared provider, otherwise one team could open itself to every stranger on Gmail. The list comes from [`free-email-domains`](https://github.com/Kikobeats/free-email-domains) (13k domains, MIT, zero deps) rather than being maintained by hand.

Its postinstall refetches the list from three remote URLs and rewrites `domains.json`, which would make the installed content vary per install, escape the lockfile, and fail the install whenever a source is down. It is therefore set to `false` in `allowBuilds`: we take the deterministic published snapshot, topped up in `public-email-domains.ts` for the ~14 providers it misses (Hey, Tutanota, Mailbox.org, some regional ISPs), with a test guarding those.

## Staff

The trial pipeline shows "Found via …" under each team name, via `TeamStaffOwner`.

## Verified

Driven locally against the test database: both redirect cases (`?r=` → that path; absent → `/`), a hostile `?r=https://evil.test` falling back to `/`, the source and skip both persisting, `team_domains` gaining the domain when checked, the checkbox appearing for a company-domain user and staying hidden for a Gmail one, and light + dark mode.

`387` unit and `316` e2e tests pass.

## Notes for review

- `structure.sql` carries incidental churn from being regenerated on a different pg_dump minor (`17.9` → `17.5`), plus an index from `20260726130000` that had not been dumped yet. The substantive part is the three `users` columns.
- Pre-existing on `main`, untouched here: team settings still lets an admin add `gmail.com` as a team domain if they have a verified Gmail address. The new eligibility rule would close that, but it could break teams that already have such a domain configured, so it is left alone.

### DIFF
--- a/apps/backend/src/util/public-email-domains.ts
+++ b/apps/backend/src/util/public-email-domains.ts
@@ -1,5 +1,3 @@
-import freeEmailDomains from "free-email-domains";
-
 /**
  * Providers `free-email-domains` does not list yet.
  *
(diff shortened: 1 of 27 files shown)


apps/backend/src/util/public-email-domains.ts
```diff
@@ -0,0 +1,51 @@
...
+const PUBLIC_EMAIL_DOMAINS: ReadonlySet<string> = new Set([
+  ...freeEmailDomains.map((domain) => domain.toLowerCase()),
+  ...EXTRA_PUBLIC_EMAIL_DOMAINS,
+]);
+
+/**
+ * Whether the domain belongs to an email provider anyone can sign up with,
+ * rather than to an organization.
+ */
+export function checkIsPublicEmailDomain(domain: string): boolean {
```
gregberge: Let's make it async and lazy-load the package if it's heavy.

PR #2412, opened 2026-07-30
feat(comments): allow commenting on a test

gregberge:
Comments could only be posted on a build. They can now also be posted on a test, so a recurring flake can be discussed where it is analysed rather than on whichever build happened to surface it.

## Backend

`comments.buildId` becomes nullable and a `testId` joins it, with a `comments_target_xor` check constraint keeping exactly one of the two set. Test comments are plain discussion threads, so `comments_test_target_scope` keeps the build-only columns (`buildReviewId`, `screenshotDiffId`) null on them.

Everything that used to take a build now takes a `CommentTarget` (`comment/target.ts`), so the comment primitives — creation, edition, deletion, reactions, thread resolution, live events, notifications — are written once and serve both. The Redis channel a change is broadcast on is derived from the comment itself, so no call site needs to know which kind it is.

The comment notification handlers name either a build or a test through a shared descriptor. Their schema fields stay optional so workflows queued before this change still validate.

**Notifications on a test root comment reach only the users it mentions**; replies then reach the thread subscribers. Builds keep their subscriber list — there is no equivalent for tests, and adding test subscriptions felt out of scope. Happy to follow up if you'd rather have it.

## API

GraphQL gains `Test.comments`, `Test.members`, the `addTestComment` mutation and the `testCommentChanged` subscription.

Every REST comment endpoint is now registered on a test path as well as a build path, sharing one implementation:

```
/projects/{owner}/{project}/tests/{testId}/comments
/projects/{owner}/{project}/tests/{testId}/comments/{commentId}
/projects/{owner}/{project}/tests/{testId}/comments/{commentId}/reactions
/projects/{owner}/{project}/tests/{testId}/comments/{commentId}/resolve
/projects/{owner}/{project}/tests/{testId}/comments/{commentId}/unresolve
/projects/{owner}/{project}/tests/{testId}/comments/{commentId}/subscription
```

`Comment.buildId` is now nullable in the public schema, next to a new `testId`. Both are additive for existing consumers; the new operations join the MCP tool surface like their build counterparts.

## Frontend

The test page grows a right column holding the change history (First change / Last change, moved out of the metrics card) and an Activity section with the comment feed and composer, kept live over the subscription.

`CommentCard` and its target-agnostic siblings move to `containers/Comment` and take how to reply plus an optional snapshot reference as props; `BuildCommentCard` wires the build side up (screenshot reference, diff navigation, build reply mutation) so the build behaviour is unchanged.

### Drive-by fix

`Editor` waited on `editor` rather than on its view being mounted, so the boxed composer threw `[tiptap error]: The editor view is not available` on first render. The build page never hit it (it renders the composer through `renderSubmit`), but the test page did. `Editor` now waits for the editor to be mounted before handing it to the toolbar or forwarding its DOM node.

## Testing

- `pnpm run static-checks` — clean
- `pnpm test:unit` — 394 passed
- `NODE_ENV=test pnpm test:integration` — 824 passed, incl. new `addTestComment.e2e.test.ts` and `testComments.e2e.test.ts`
- `NODE_ENV=test pnpm test:e2e` — 70 passed, incl. a new spec posting a comment from the test page
- Migration verified up **and** down against a local database


apps/backend/src/api/handlers/addCommentReaction.ts
```diff
@@ -21,79 +24,131 @@ import {
...
+  },
+  "400": invalidParameters,
+  "401": unauthorized,
+  "403": forbidden,
+  "404": notFound,
+  "500": serverError,
+};
+
 export const addCommentReactionOperation = {
   operationId: "addCommentReaction",
```
gregberge: We should rename it to "addBuildCommentReaction". Same things for other endpoints.

apps/backend/src/comment/createComment.ts
```diff
@@ -140,39 +154,68 @@ export async function createBuildComment(input: {
...
+}): Promise<void> {
+  const { target, userId } = input;
+  switch (target.type) {
+    case "build":
+      await autoSubscribeUserToBuild({ buildId: target.build.id, userId });
+      return;
+    case "test":
+      await autoSubscribeUserToTest({ testId: target.test.id, userId });
+      return;
+  }
```
gregberge, line 192: assertNever?

apps/backend/src/comment/createComment.ts
```diff
@@ -140,39 +154,68 @@ export async function createBuildComment(input: {
...
+/** The users following a build or a test. */
+async function getTargetSubscribedUserIds(
+  target: CommentTarget,
+): Promise<string[]> {
+  switch (target.type) {
+    case "build":
+      return getBuildSubscribedUserIds(target.build.id);
+    case "test":
+      return getTestSubscribedUserIds(target.test.id);
+  }
```
gregberge, line 206: assertNever?

apps/backend/src/comment/getVisibleComments.ts
```diff
@@ -45,3 +47,39 @@ export async function getVisibleBuildComments(input: {
...
+  const { target, viewerUserId } = input;
+  switch (target.type) {
+    case "build":
+      return getVisibleBuildComments({
+        buildId: target.build.id,
+        viewerUserId,
+      });
+    case "test":
+      return getVisibleTestComments({ testId: target.test.id, viewerUserId });
+  }
```
gregberge, line 123: assertNever?

apps/backend/src/comment/target.ts
```diff
@@ -0,0 +1,126 @@
...
+export function getCommentTargetColumns(target: CommentTarget): {
+  buildId: string | null;
+  testId: string | null;
+} {
+  switch (target.type) {
+    case "build":
+      return { buildId: target.build.id, testId: null };
+    case "test":
+      return { buildId: null, testId: target.test.id };
+  }
```
gregberge, line 30: assertNever, check also others.

apps/backend/src/comment/target.ts
```diff
@@ -0,0 +1,126 @@
...
+    }
+    case "test": {
+      const { test } = target;
+      if (!test.project?.account) {
+        await test.$fetchGraph("project.account");
+      }
+      invariant(test.project?.account, "Test project account not found");
+      return test.project;
+    }
+  }
```
gregberge, line 74: assertNever

apps/backend/src/comment/target.ts
```diff
@@ -0,0 +1,126 @@
...
+/** URL of the page a target's comments are shown on. */
+export async function getCommentTargetUrl(
+  target: CommentTarget,
+): Promise<string> {
+  switch (target.type) {
+    case "build":
+      return target.build.getUrl();
+    case "test":
+      return target.test.getUrl();
+  }
```
gregberge, line 111: assertNever

apps/backend/src/comment/target.ts
```diff
@@ -0,0 +1,126 @@
...
+} {
+  switch (target.type) {
+    case "build":
+      return {
+        buildNumber: target.build.number,
+        buildName: target.build.name,
+      };
+    case "test":
+      return { testName: target.test.name };
+  }
```
gregberge, line 134: assertNever

apps/frontend/src/pages/Test/ActivitySection.tsx
```diff
@@ -0,0 +1,390 @@
...
+    }
+  }
+`);
+
+/**
+ * Reads a `#comment-…` hash from the URL and, when it matches a loaded comment,
+ * returns its id so the comment can be highlighted. The highlight clears on the
+ * next click anywhere, after 3 seconds, or when the component unmounts.
+ */
+function useHighlightedCommentId(commentIds: string[]): string | null {
```
gregberge: Can it be factorized because I think we have the same logic on the build?

apps/backend/src/api/handlers/addCommentReaction.ts
gregberge: Done — renamed all eleven build comment operations to carry the target: `addBuildCommentReaction`, `listBuildComments`, `createBuildComment`, `getBuildComment`, `updateBuildComment`, `deleteBuildComment`, `removeBuildCommentReaction`, `resolve`/`unresolveBuildCommentThread`, `subscribe`/`unsubscribeBuildCommentThread`.

Worth knowing: operation ids are the public names — they become the MCP tool names (snapshot updated in `tools.test.ts`) and would name generated SDK methods, so this is a breaking rename for anything calling the old tool names. The URLs are untouched. Happy to keep aliases if that matters.

apps/backend/src/comment/createComment.ts
gregberge: Done.

apps/backend/src/comment/createComment.ts
gregberge: Done.

apps/backend/src/comment/getVisibleComments.ts
gregberge: Done.

apps/backend/src/comment/target.ts
gregberge: Done — all four switches in this file, plus the two in `createComment.ts` and the one in `getVisibleComments.ts`. Seven in total, every switch over `target.type`.

I left the `switch (event.type)` in the two activity sections alone: those deliberately handle only Added/Deleted (Updated needs no cache work), so they are not meant to be exhaustive.

apps/backend/src/comment/target.ts
gregberge: Done.

apps/backend/src/comment/target.ts
gregberge: Done.

apps/backend/src/comment/target.ts
gregberge: Done.

apps/frontend/src/pages/Test/ActivitySection.tsx
gregberge: Good catch — the two copies were byte-identical. Extracted to `containers/Comment/useHighlightedCommentId.ts` and both feeds now import it.

PR #2423, opened 2026-08-03
Drop the Claude Code terminal deep link

jsfez:
`claude-cli://` opens a terminal, and on macOS it does so by having AppleScript type its launch command into iTerm2 or Terminal.app. A typed line is cut at 1,024 bytes by the tty, and the prompt rides there base64-encoded, so anything past roughly 650 characters reached the agent truncated mid-sentence — silently, since a cut base64 tail still decodes.

`claude://code/new` opens the same Claude Code session inside the desktop app, carries around 14,000 characters and never goes near a terminal. It becomes the Claude entry, and the menu now reads "Open in Claude".

Prompts are untouched. Anyone who had picked Claude Code falls back to copying, which `AiPromptButton` already handles for a target that no longer exists.

### DIFF
--- a/apps/frontend/src/containers/agent-icons.tsx
+++ b/apps/frontend/src/containers/agent-icons.tsx
@@ -27,24 +27,6 @@ export function ClaudeIcon({ className }: IconProps) {
   );
 }
 
-export function ClaudeCodeIcon({ className }: IconProps) {
-  return (
-    <svg
-      className={className}
-      viewBox="0 0 24 24"
-      aria-hidden="true"
-      fill="currentColor"
-      fillRule="evenodd"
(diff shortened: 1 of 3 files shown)


gregberge (dismissed):
Can't we find another solution? I don't want to degrade our prompts for this. I would rather prefer to remove Claude Code Terminal in this case.

jsfez:
> Can't we find another solution? I don't want to degrade our prompts for this. I would rather prefer to remove Claude Code Terminal in this case.

We didn't degrade the prompt quality. We rely on the skill to keep the prompt concise.
Another option would be to remove the Claude Code deep link and rely only on "Copy prompt."

gregberge:
@jsfez yes I prefer to remove Claude Code in this case

PR #2425, opened 2026-08-04
feat(review): optionally return to the pull request after review

jsfez:
Adds a checkbox at the bottom of the review panel — shown only when the build is linked to a pull request — to automatically navigate back to the PR after submitting a review. The preference is persisted in localStorage, so once checked it applies to subsequent reviews.

### DIFF
--- a/apps/frontend/src/pages/Build/BuildReviewForm.tsx
+++ b/apps/frontend/src/pages/Build/BuildReviewForm.tsx
@@ -144,11 +144,7 @@ export function BuildReviewForm(props: {
       // The mutation closes the dialog through `onCompleted`; keep the pending
       // state until then so the form doesn't flash back to idle and the modal
       // stays locked through the closing animation.
-      if (
-        returnToPullRequest &&
-        pullRequestUrl &&
-        event !== BuildReviewEvent.Comment
-      ) {
+      if (returnToPullRequest && pullRequestUrl) {
         window.location.assign(pullRequestUrl);
       }
     } catch (error) {
@@ -226,7 +222,7 @@ export function BuildReviewForm(props: {
             }}
             className="text-low self-start text-sm"
           >
-            Return to the pull request after approving or rejecting
+            Return to the pull request after review
           </Checkbox>
         ) : null}
         <Tooltip
--- a/apps/frontend/src/pages/Build/BuildReviewForm.tsx
+++ b/apps/frontend/src/pages/Build/BuildReviewForm.tsx
@@ -6,6 +6,7 @@ import { useForm } from "react-hook-form";
 import { DocumentType, graphql } from "@/gql";
 import { BuildReviewEvent } from "@/gql/graphql";
 import { Button, ButtonIcon } from "@/ui/Button";
+import { Checkbox } from "@/ui/Checkbox";
 import { useOverlayTriggerState } from "@/ui/Dialog";
 import { type EditorValue } from "@/ui/Editor/Editor";
 import { EditorField } from "@/ui/Editor/EditorField";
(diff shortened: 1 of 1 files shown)


apps/frontend/src/pages/Build/BuildReviewForm.tsx
```diff
@@ -131,6 +144,13 @@ export function BuildReviewForm(props: {
       // The mutation closes the dialog through `onCompleted`; keep the pending
       // state until then so the form doesn't flash back to idle and the modal
       // stays locked through the closing animation.
+      if (
+        returnToPullRequest &&
+        pullRequestUrl &&
+        event !== BuildReviewEvent.Comment
```
gregberge: Why not for comment? It's disturbing.

apps/frontend/src/pages/Build/BuildReviewForm.tsx
gregberge, line 1: I am not sure to see the interest, it only saves one click because the PR is accessible from the header. If it's because of keyboard access, then a shortcut to go to the PR is more useful. So I prefer to go this way because if at some point you decide to not review it you can still go back anytime to the PR.

gregberge (changes requested):

apps/frontend/src/pages/Build/BuildReviewForm.tsx
jsfez: I agree with you

apps/frontend/src/pages/Build/BuildReviewForm.tsx
jsfez: The goal of this feature isn't to save a click, it's to remove the "what do I do now?" moment after finishing a review. It's about reducing friction and giving users a clear next step.

A keyboard shortcut would definitely be useful, but mostly for power users. Today, I suspect only a small fraction of users even know they can click the PR link in the header to go back.

gregberge: I understand but I have never heard this feedback from any of our users so I am not sure if it exists or not. If the user opened the link from GitHub in a new tab, it ends up with two tabs at the end and it's a bit confusing. People are very familiar with checking the result of tests on GitHub and going back to the PR, this flow is already known by everyone.

jsfez: > I have never heard this feedback from any of our users so I am not sure if it exists or not.

I'm not too surprised. This is the kind of small UX friction that users usually don't report because they've already adapted to it. It's something you notice more by observing users than by asking them.

I can at least say that I've personally felt this friction several times while using Argos.

> If the user opened the link from GitHub in a new tab, it ends up with two tabs at the end and it's a bit confusing.

I don't think that's a significant issue. If users end up with two tabs, they'll most likely just close one.

> People are very familiar with checking the result of tests on GitHub and going back to the PR, this flow is already known by everyone.

I actually find that flow a bit annoying on GitHub too. Just because it's familiar doesn't necessarily mean it's the best experience. If we can remove that friction without cluttering the UI, I think it's worth doing.

I checked out the PR to test it. The checkbox doesn't clutter the UI, and I think it's a good addition.

CleanShot 2026-08-04 at 17 45 55@2x

gregberge: I think it clutters the UI a bit to be honest, it creates two rows of actions. IMO it's harder to understand now because it's a bit "out of the blue".

gregberge: I don't want to ship this in this state, in terms of design it's not clean.

jsfez: They are several way to handle this topic and I'm not stuck on this solution. 
Let's close this PR and discuss it later

PR #2426, opened 2026-08-04
feat(build): add onion skin and swipe diff view modes

jsfez:
Adds two view modes for comparable image diffs, alongside Baseline and Changes: onion skin fades the changes screenshot over the baseline with a floating opacity slider, and swipe reveals it with a draggable divider that keeps a constant on-screen size at any zoom. Bound to the O and W hotkeys, with a fallback to split view for diffs that cannot be blended (added, removed, text snapshots). The changes overlay stays available in both modes: shown over the full blend in onion skin, clipped to the changes side of the divider in swipe.

### DIFF
--- a/apps/frontend/src/containers/Build/BuildDiffDetail.tsx
+++ b/apps/frontend/src/containers/Build/BuildDiffDetail.tsx
@@ -14,10 +14,11 @@ import { assertNever } from "@argos/util/assertNever";
 import { invariant } from "@argos/util/invariant";
 import { captureException } from "@sentry/react";
 import { clsx } from "clsx";
-import { useAtomValue } from "jotai/react";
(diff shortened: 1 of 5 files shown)


apps/frontend/src/containers/Build/BuildDiffDetail.tsx
gregberge, line 1: We can't set the opacity to 0, it's a bit weird, maybe we should revisit the onion, but else it's very good!

gregberge (approved):

apps/frontend/src/containers/Build/BuildDiffDetail.tsx
jsfez: You can set the opacity to 0 if you hide the diff layer (red eye button)

PR #2430, opened 2026-08-04
perf(frontend): flatten the boot waterfall, and harden the app's security headers

gregberge:
Flattens the frontend's boot waterfall and shrinks what it downloads, then fixes
the security headers on the app domain. One commit per change, each independently
revertable.

## Perf

Boot critical path: **2.35 MB → 1.31 MB raw, 509 kB → 409 kB gzip**. First paint
went from four serial GraphQL round trips to one.

The cascade was `Auth_me` → `FeatureFlagProvider_account` → `Project_project` →
`ProjectBuilds_*`. The first two were pure gates: one rendered a `PageLoader`,
the other rendered `null`, each holding back everything below it. Because these
routes have no loaders — all fetching is render-driven — no page could even *ask*
for its data until both resolved.

| Commit | Effect |
| --- | --- |
| Remove the unused Reflag system | 168 kB SDK off the entry chunk, one serial round trip gone |
| Render the app before the user resolves | Page queries start on first paint instead of a round trip later |
| Tree-shake GraphQL documents per route | 940 kB chunk eliminated; 176 → 17 operation ASTs on the critical path |
| Load the editor and emoji picker on demand | Build page static graph 1335 kB → 317 kB (306 → 87 kB gzip) |
| Serve the navbar avatar from `Auth_me` | One round trip gone |
| Match `manualChunks` on package names | `react` chunk 474 → 178 kB; react-hook-form, react-redux, recharts off boot |
| Take fonts and runtime config off the critical path | Two third-party origins and one origin round trip removed |

### Notes

**Reflag was never used.** Both declared flags were vestigial — the features are
gated on GraphQL fields (`ignoreConfig.enabled`, `deploymentEnabled`) — and
`featureGuardHoc`, the only consumer of `useFlag`, had no call sites.

**Rendering before the user resolves** cost less than expected:
`useAssertAuthTokenPayload` had only four call sites, all leaf member/contributor
lists already inside Suspense boundaries. It now suspends on `me` instead of
reading context; Apollo deduplicates against the provider's in-flight request, so
no extra round trip. `useIsLoggedIn` trusts the `argos_logged_in` hint while `me`
is in flight, and the navbar reserves the avatar's footprint, so nothing flashes
or reflows.

**GraphQL documents** needed a new plugin. The codegen client preset resolves
documents through a map keyed by query string, which names every operation in the
app — the generated file says so itself: *"it is not tree-shakeable, so it will
include all operations in the project"*. Codegen ships Babel and SWC plugins for
exactly this, but neither applies: `@vitejs/plugin-react` v6 on Vite 8 transforms
with Oxc. So `@argos/vite-plugin-graphql-documents` does the same rewrite as a
Vite transform, with a README and 26 unit tests.

Worth knowing *why* it inlines the AST rather than importing the generated
constant: `graphql.ts` is a single module, so while chunks import documents *from*
it, Rolldown must emit it as one shared chunk holding every document any chunk
needs. That variant still left all 176 operations on the critical path. Inlining
is what actually splits them. Each source string occurs exactly once in the
codebase, so nothing is duplicated. Unknown literals fall back to the runtime
lookup with a warning, so stale codegen degrades rather than breaking the build.

**`manualChunks` was misgrouping.** `patterns.some(p => id.includes(p))` meant
`node_modules/react` also matched `react-hook-form`, `react-router` and
`react-redux`, so the boot chunk shipped a form library and — via recharts'
dependency on `@reduxjs/toolkit` — Redux, to pages that render neither. It now
resolves the package name from the last `node_modules/` segment, which matters
under pnpm, whose directory names embed peer versions
(`react-dom@19.2.8_react@19.2.8`). Also drops the dead `moment` and `lodash`
patterns.

**Fonts and config.** Inter came from `fonts.googleapis.com`, which then
referenced `fonts.gstatic.com` — two DNS+TLS handshakes before any text could
paint. Now self-hosted as a weight-axis variable font: one same-origin,
content-hashed file. `/config.js` was a parser-blocking, uncacheable request
sitting ahead of the entry bundle that cannot boot without it; production now
inlines it into the shell, with a CSP hash computed once at startup.

### Not done

`BatchHttpLink` is deliberately left out, for two reasons worth deciding on
separately:

1. It needs `allowBatchedHttpRequests: true` on Apollo Server, and the `/graphql`
   rate limiter counts per *request* — a batch of N operations would count as 1,
   weakening it.
2. It introduces head-of-line blocking. On first paint `Auth_me`, the breadcrumb
   query and `Project_project` all fire in the same tick, so the navbar would wait
   for the slowest — undoing the commit that makes the shell render immediately.

The duplicate-request half of that idea is done: the avatar fold removes the
query that existed only to fetch an avatar for an account `Auth_me` had already
resolved.

## Security headers

Four problems on the app domain's helmet block, all verified against live headers
on `app.argos-ci.com`. Extracted into `security-headers.ts` so the policy can be
asserted on directly.

- **`connect-src` was `["'self'", "*"]`.** The `*` cancelled the `'self'` beside
  it and left the directive meaningless: any XSS or compromised dependency could
  exfiltrate anywhere. Now an allowlist derived from config, so it cannot drift
  from where the app actually talks — including the image CDN and signed-S3
  origins that screenshots and text snapshots are *fetched* from, not just
  displayed.
- **`frame-ancestors` was absent**, so helmet's `'self'` default applied. Browsers
  prefer CSP over X-Frame-Options, so same-origin framing was in fact allowed and
  the `frameguard: deny` beside it was dead weight.
- **COOP was off entirely**, leaving `window.opener` intact for cross-origin
  openers (tabnabbing / XS-Leaks). Now `same-origin-allow-popups`.
- **`style-src` and `font-src` inherited helmet's blanket `https:`.** Narrowed;
  fonts are self-hosted now, so no third-party origin is needed.

Plus a `Permissions-Policy` denying camera, microphone, geolocation and payment,
and deliberate cache headers (`no-cache` + `Vary` on the shell and `/config.js`,
`immutable` on content-hashed assets).

### Verification

Because this CSP only applies when the **backend** serves the built SPA — not via
the Vite dev server — it was exercised in a real browser in that mode, across
login, home, account, project, a build with screenshots, and both settings pages:
**zero CSP violations**, with requests to `files.argos-ci.com` and the S3 bucket
both succeeding. Covered by 16 unit tests, including that `connect-src` never
regains a wildcard and never leaks the Sentry DSN's credentials.

All 26 static checks and 549 unit tests pass.

### Needs your input

- **CORP is still off.** I could not verify whether `argos-ci.com` or the docs
  site embeds assets from `app.argos-ci.com`, so I left it with a comment rather
  than guessing.
- **HSTS `preload`** is flagged, not added — only safe if every subdomain is
  HTTPS-only.
- **CSP reporting stays unconfigured**, per the decision that the reports are
  noise we cannot action.
- **Latent, not changed:** `@shikijs/engine-oniguruma` calls
  `WebAssembly.instantiate`, which would need `script-src 'wasm-unsafe-eval'`.
  Inert today — only `createJavaScriptRegexEngine` is actually invoked — but
  switching shiki's regex engine would break code diffs under CSP.


gregberge:
Addressed both.

**CodeQL** — fair catch: `origin.includes("amazonaws.com")` would also pass for `https://amazonaws.com.attacker.test`, so it proved little. The test now parses each entry and matches the whole host, anchored (b0e4d89f).

**Broken e2e (`staff all teams`, `staff trial pipeline`)** — a real bug I introduced, not flaky. Rendering the app before `me` resolves left `AuthGuard` seeing a null account on the first render and reading it as *signed out*, so every guarded route (`/`, `/new`, `/teams/new`, `/staff/*`) redirected a signed-in user to `/login`. The chain then recovered — `/login` forwarded them to their account — which is why it showed up as the staff pages failing rather than as an obvious breakage.

Reproduced against the built app before fixing:

```
/            -> / -> /login -> /login -> / -> /csp-user
/new         -> /new -> /login -> /login -> / -> /csp-user
/teams/new   -> /teams/new -> /login -> /login -> / -> /csp-user
/staff/teams -> /staff/teams -> /login -> /login -> / -> /csp-user   <- never arrives
```

and after:

```
/            -> / -> /csp-user
/new         -> /new -> /csp-user/new
/teams/new   -> /teams/new
/staff/teams -> /staff/teams
```

The root cause was that `useIsLoggedIn` returned a boolean and so could not express "not known yet". It's replaced by `useAuthStatus` → `"loading" | "authenticated" | "anonymous"`, and **removed** rather than kept alongside, so the ambiguous version isn't reachable. All twelve consumers are converted; the seven that pick a destination now hold on `"loading"` (several are mid-flow — CLI login, OAuth consent — where a stray redirect drops the parameters the page was reached with), and the rest withhold an affordance, which is the safe direction.

Also verified the guard still works: signed-out visitors are redirected, and a stale `argos_logged_in` hint settles on `/login` without looping. `tests/auth-guard.spec.ts` covers all of it, asserting on the *navigation chain* — the final URL looked reasonable even while the bug was present, so that's what makes it catchable.

Still needs your call, unchanged from the description: the CORP decision, HSTS `preload`, and whether to add `BatchHttpLink`.

jsfez (approved):

PR #2445, opened 2026-08-08
feat(media): upload standalone images and videos

gregberge:
Implements standalone media upload: an image or video that belongs to no build and no test run, with a shareable URL and ready-to-paste Markdown.

The primary caller is an AI coding agent. GitHub has no public API for comment attachments, so an agent working from a terminal has no other way to put a screenshot in the pull request it just opened.

**Draft — iterating.** Companion PRs: [argos-javascript](https://github.com/argos-ci/argos-javascript) (CLI, SDK, agent skill), [docs](https://github.com/argos-ci/docs), [argos-ci.com](https://github.com/argos-ci/argos-ci.com) (pricing, changelog).

## The shape

A **media** is an identity — what the picture is _of_ — and holds no bytes. Each upload is a **version** of it. That is what makes re-uploading after review work: the share URL keeps pointing at the newest image, and the version a reviewer commented on survives underneath it. Identity is `(project, what it is attached to, name, state)`, so the same name uploaded again stacks rather than collides.

It is scoped to a **project**, like a build — so it inherits project permissions, project transfer, project deletion, and bills through the project's account.

### Attached to a branch, or to a pull request

An agent produces screenshots while it is still doing the work; the pull request comes after, if it comes at all. So a media attaches to a **branch** (`stage: "staged"`) or to a **pull request** (`stage: "published"`).

A staged media is real — shareable, listed, with its URL from the moment it is registered, and no GitHub connection needed to name a branch. Opening the pull request for that branch is what publishes it: Argos attaches everything staged there and posts its comment, unasked. Nothing has to come back and connect the two.

A media keeps its branch after publishing, as provenance, but the branch stops counting towards identity once a pull request is attached — otherwise re-uploading with `prNumber` would read as a different media.

### One comment per pull request

Argos maintains a single comment listing every media uploaded, editing it in place. Attaching a media to a pull request and showing it there are the same act, not two. Videos embed their poster frame linked to the share page, because GitHub only renders inline players for media it hosts itself.

## Surface

**REST** (new `media:read` / `media:write` scopes; the MCP tools derive from these automatically):

|                                         |                                                                          |
| :-------------------------------------- | :----------------------------------------------------------------------- |
| `POST /media` (+ `/finalize`)           | register and confirm an upload                                           |
| `GET` `PATCH` `DELETE /media/{id}` | read, edit while staged, remove |
| `GET /media/{id}/versions` | the upload history, when `versionCount` says there is one |
| `GET /projects/{owner}/{project}/media` | filter by `branch`, `prNumber`, `stage`, `search`, `type`                |
| `/media/{id}/comments/…`                | the full comment surface — threads, reactions, resolution, subscriptions |

`PATCH` applies while staged only. Once published, the name is what the pull request comment is built from and what review threads hang off, so editing it rewrites history rather than correcting a staged media.

A comment carries the `mediaVersionId` it was written against. A pin describes a spot on the bytes its author was looking at, so acting on feedback means reaching _that_ file — which `GET /media/{id}/versions` resolves. It is a separate call because it is rarely needed: most media have one version, and `versionCount` on the media says whether to bother.

**Share page** at `/m/:token`, outside the app layout, built on the build viewer's grammar: pan/zoom, before-after comparison, pinned comments, version history, hotkeys. `visibility` decides who may open it — `team` by default, so a private repository's screenshots don't become world-readable by being uploaded.

**Billing**: the existing screenshot meter, 1 unit per image and 25 per video, so it lands on the invoice line accounts already read and inherits the capacity and spend-limit checks unchanged. The usage detail breaks media out — a single video is 25 units, and four recordings looking like a hundred unexplained screenshots is not a usable bill.

**Retention** purges per version, hourly, bytes included: an old version ages out while the media and its share URL live on.

## No server-side processing

Argos never rewrites an uploaded file. The bytes are stored as given and served from ImageKit, which derives WebP/AVIF variants and video poster frames on request.

Finalizing does **one 64 KB ranged read** — `file-type` identifies the container, sharp reads an image's dimensions off the header. That is the whole of the server-side handling, it runs inline, and a media is usable the moment the upload finishes.

An earlier revision had an ffmpeg worker (poster extraction, transcode) and a sharp image re-encode. Measured on a real 1536×1024 Argos screenshot, the lossless WebP re-encode cost **711 ms of CPU to save 35% of bytes** — work ImageKit already does better at request time. Both were removed.

## Two things to review deliberately

**Files are served unauthenticated**, from an unguessable content-addressed CDN URL, whatever the visibility. This is forced by the purpose: GitHub renders an embedded image by fetching it server-side through camo, with no Argos session, so an authenticated URL cannot be embedded at all. `visibility` therefore governs the **share page** (project context, expiry, delete), not the bytes. Same posture Argos already uses for build screenshots.

That unauthenticated URL is also why the content check matters: a caller can declare `image/png` and upload HTML, and nothing else looks at the bytes. Identifying them is what keeps active content off an Argos domain.

**Embedded metadata is preserved.** Since nothing rewrites the file, a photo's EXIF — including GPS — stays in the stored original. The CDN drops it from the variants a browser fetches, but not from the original. Documented rather than silently dropped; stripping it client-side in the CLI is a possible follow-up.

## Verification

`pnpm run static-checks` (all 18 CI tasks), 584 unit, 1114 integration, Playwright specs — all green.

One assumption I could not verify locally: that ImageKit's `ik-thumbnail.jpg` resolves for video on our plan and URL endpoint. **Worth hitting one real video URL before this leaves draft** — the video embed rests on it.

## Not in this PR

- **No listing UI.** Media is reached by its link. The account-level "Pull requests" tab that listed every pull request with media on it was removed — a browsing surface nobody has asked for yet, and a lot of machinery to carry. It can come back when there is a reason for it to exist, and it will want a different shape by then.
- `media/serve.ts` uses `getTwicPicsUrl`, the helper name on `main`. #TBD renames it to `getImageKitUrl`; whichever merges second needs the one-line follow-up.


apps/backend/src/api/handlers/listMediaFeedback.ts
```diff
@@ -0,0 +1,260 @@
...
+/**
+ * How many media a single feedback fetch covers. A pull request with more
+ * uploaded media than this has other problems, and an unbounded fan-out here
+ * would let one request read a project's whole history.
+ */
+const MAX_MEDIA = 100;
+
+const MediaFeedbackSchema = z
+  .object({
+    media: z
```
gregberge: Don't we have a shared media schema?

apps/backend/src/api/handlers/listMediaFeedback.ts
gregberge, line 1: Actually I don't understand the usage of this endpoint, we should remove it.

apps/backend/src/api/handlers/media.ts
```diff
@@ -0,0 +1,541 @@
...
+    .meta({
+      description:
+        "Project to upload to, as `owner/project`. Required with a personal access token; ignored with a project token, which already identifies its project.",
+      examples: ["acme/web"],
+    }),
+  prNumber: z.number().int().min(1).nullish().meta({
+    description:
+      "Pull request this media belongs to. Argos maintains a single comment on it listing every media uploaded, editing it in place rather than posting a new one each time — attaching a media to a pull request and showing it there are the same act, not two. Also part of the media's identity: uploading the same name again on this pull request adds a version.",
+  }),
+});
```
gregberge, line 99: An agent should be able to create some medias before the pull request is created. To do this, it should be able to reference the media from a branch. In this case once the PR is opened, the comment is automatically added.

We should also add an endpoint to list medias attached to a PR or attached to a branch. Those linked to a branch should be in a stage "draft" not yet published.

We should also add an API to edit them (description, name, branch) when it's still "staged". Once published the PR number is attached and they are not yet "staged".

apps/backend/src/api/handlers/media.ts
```diff
@@ -0,0 +1,541 @@
...
+    }),
+    query: PageParamsSchema.extend({
+      search: z.string().optional().meta({
+        description: "Match media on their file name or slug.",
+      }),
+      type: z.enum(["image", "video"]).optional().meta({
+        description: "Restrict to images or to videos.",
+      }),
+    }),
+  },
```
gregberge, line 485: What is important here, is filtering by a branch or a PR number.

apps/backend/src/api/handlers/media.ts
```diff
@@ -0,0 +1,541 @@
...
+    // The comment lists what is still there, so it has to lose this entry.
+    if (media.githubPullRequestId) {
+      await updatePullRequestComment(media.githubPullRequestId);
+    }
+
+    res.status(204).send();
+  });
+};
+
+export const listMediaOperation = {
```
gregberge, line 440: Think about a smart way of giving access to the previous versions of a media without bloating the API. Because a comment can be posted on a previous version and we should still be able to access the file of this version from the API. Maybe we could just list all of them directly in the media?

apps/backend/src/database/services/period-usage.ts
gregberge, line 1: Make sure to update the UI in the account settings in the billing to display the usage of medias in the detail also. It's important for the user to understand how many medias has been uploaded in the period.

apps/backend/src/graphql/definitions/Account.ts
```diff
@@ -176,6 +177,14 @@ export const typeDefs = gql`
       period: MetricsPeriod!
       filters: TestsFilterInput
     ): TestConnection!
+    """
+    Pull requests with media uploaded to them, across the projects the viewer can
+    see, most recent first.
+    """
+    mediaPullRequests(
```
gregberge: Finally get rid of the pull requests view in this PR. It's too much, let's just focus on medias. For now medias will be only accessible for people with the link, they will be no UI to list them.

apps/backend/src/media/sniff.ts
gregberge, line 1: Do we really need this? Can't we rely on a library, also why are we doing this?

apps/backend/src/media/purge.ts
gregberge, line 1: The purge of medias, should also purge them from S3 of course.

apps/backend/src/api/handlers/listMediaFeedback.ts
gregberge: Removed in 39302b51.

You are right that it had no job of its own. `listMediaComments` already returns a media's comments through the shared `Comment` schema, and the media is in hand by the time you ask — the grouping was the only thing this added, and grouping is the caller's business.

What it *was* covering is now covered properly: a comment carries `mediaId` and `mediaVersionId` (177c63cf), so feedback says which media and which upload it is about without a bespoke endpoint to tell you.

apps/backend/src/api/handlers/listMediaFeedback.ts
gregberge: Moot now — the endpoint is gone in 39302b51.

But the answer was no, and that was the tell: this was a second, thinner `Media` inlined into one response schema, which would have had to be kept in step with the real one forever. Everything that reads a media now goes through `MediaSchema` in `apps/backend/src/api/schema/primitives/media.ts`.

apps/backend/src/graphql/definitions/Account.ts
gregberge: Done in a4cb5f75 — the tab, the page, the row component, the route, the `mediaPullRequests` field on all three account shapes, its connection type, two dataloaders, the aggregate query and two visual tests.

None of it was load-bearing for uploading, sharing a link, or the pull request comment. Removed rather than kept warm: it can come back when there is a reason for it to exist, and it will want a different shape by then.

apps/backend/src/media/purge.ts
gregberge: It already did — but nothing proved it, which is the real problem, so 899037d6 adds the test.

Every path that removes a media goes through `deleteUnreferencedMediaObjects`, the only caller of the private S3 delete: retention, an explicit delete, a rejected upload, and project (so account) deletion. The reason this needed a test rather than a fix is that it is a silent failure mode — rows vanish, share URLs stop working, everything looks healthy, and the bytes stay in the bucket forever.

The covered cases are the ones that make it non-obvious: a key a surviving version still points at must *not* be deleted (keys are content-addressed, so re-uploading an earlier screenshot makes two versions share one object), and an abandoned two-step upload still ages out — `expiresAt` is stamped at creation, not at upload, which is what makes those orphans collectable at all.

apps/backend/src/media/sniff.ts
gregberge: Replaced with `file-type` in 74df80e8.

**Why it exists at all:** it is the only point where Argos looks at an uploaded file. The API validates the content type the *caller declares* and S3 enforces the upload matches that declaration — but a caller can declare `image/png` and upload HTML. Media is served from an unauthenticated CDN URL, so that would host attacker-controlled active content on an Argos domain.

**Why hand-rolling it was wrong:** the tail is where the bypass lives. Our table read the EBML magic bytes and answered `video/webm` — which every Matroska file also starts with. `file-type` reads the `DocType`, so Matroska declared as WebM is now rejected instead of stored as a video no browser will play. It also names what it rejected, so a WAV passing as WebP reads as `audio/wav` rather than as nothing.

apps/backend/src/database/services/period-usage.ts
gregberge: Done in 1d715cd4.

You were right that the number was unreadable: a single video is 25 units, so four recordings looked like a hundred screenshots nobody could account for, and the page offered no way to find out.

`ScreenshotsCount` now carries a `media` breakdown — how many uploads, and what they cost — alongside the Storybook split, filled from the same query rather than a new one. The usage detail spends it per project *and* in the total:

> 36 screenshots  (10 classic + 26 from 2 media)

The parts add up, which is why the classic figure has the media units taken back out of it. Counted per **version**, because that is what gets billed — every version stores its own bytes, so a re-upload after a fix is a new upload.

apps/backend/src/api/handlers/media.ts
gregberge: Listed directly in the media, as you suggested — 177c63cf.

Two halves were missing, not one. Versions are now on the media inline (`versions`, newest first, each with its own `fileUrl` and `posterUrl`), **and** a comment carries `mediaId` + `mediaVersionId` — it did not even say which media it was on before. So resolving feedback to the file it is actually about is a lookup in a response the caller already has.

On the bloat: the round trip is the cost, not the fields. One query per page loads them, the same as before, and `versionCount` is gone since the list answers it. Unfinalized versions stay out — those rows exist to sign an upload, and listing one hands out a URL for bytes that are not there.

apps/backend/src/api/handlers/media.ts
gregberge: Built in 47ef65ed — all four parts.

**Upload against a branch.** `POST /media` takes `branch`. The media is `stage: "draft"` — real, shareable, listed, with its URL immediately, and no GitHub connection needed to name a branch.

**The pull request publishes it.** Opening the pull request for that branch attaches every draft waiting there and posts the comment, in one pass, unasked. This needed a column Argos never had: `github_pull_requests.headRef`, the branch a pull request is *from*. Only `baseRef` was stored, and you cannot match drafts on the branch being merged into.

**List by branch or pull request.** `GET /projects/{owner}/{project}/media?branch=…&prNumber=…&stage=…`.

**Edit while draft.** `PATCH /media/{mediaId}` takes name, description and branch, and refuses once published — by then the name is what the pull request comment is built from and what review threads hang off, so editing rewrites history rather than correcting a draft. A colliding rename answers 409, not 500.

One design note worth your eye: **identity** is now `(project, whatever it is attached to, name, state)` — the pull request when there is one, the branch otherwise. A draft keeps its branch after publishing as provenance, but the branch stops counting the moment a pull request is attached. Without that, re-uploading with `prNumber` would read as a different media and create a second one instead of a version.

I went with `draft` / `published` for the stage values — you wrote both "draft" and "staged"; say the word if you prefer `staged`.

Publishing is idempotent and scoped to the pull request's own repository — branch names are not unique across an installation, and matching on the name alone would publish one repository's work onto another's pull request.

apps/backend/src/api/handlers/media.ts
gregberge: Both added in 47ef65ed: `?branch=` and `?prNumber=`, plus `?stage=` for drafts vs published.

`branch` is the one that earns its keep — a media keeps its branch after publishing, so "everything for this work" stays a single query across the moment the pull request opens, instead of the caller having to switch filters when it does.

`prNumber` resolves to the pull request of *this* project's repository rather than joining on the raw number, so it cannot reach another repository's pull request, and a number Argos has never seen answers an empty list rather than an error.

apps/backend/src/media/url.ts
```diff
@@ -0,0 +1,109 @@
...
+  return [
+    `| ${headers.join(" | ")} |`,
+    `| ${headers.map(() => "---").join(" | ")} |`,
+    ...rows,
+  ].join("\n");
+}
+
+/** `|` alone is enough to break out of a table cell. */
+function escapeTableCell(value: string): string {
+  return value.replace(/\|/g, "\\|");
```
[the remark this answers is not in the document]
gregberge: Fixed in b882cffd.

Real bug: escaping `|` writes backslashes, so a value already ending in one had our escape escaped and let the pipe through — the exact break the function exists to prevent. Backslashes are now escaped first.

Two neighbouring holes went with it, since they are the same class:
- A newline in a description ended the table row outright and has no escape, so it becomes the `<br>` GitHub renders a line break with.
- A pipe in a file name still opened a column from inside the embed's alt text, which `escapeTableCell` cannot reach — by then the label is glued to a URL that must not be escaped — so `escapeMarkdownText` escapes it at the label instead.

apps/backend/src/media/pull-request-comment.ts
```diff
@@ -0,0 +1,173 @@
...
+    "| --- | --- |",
+    ...entries,
+    "",
+    "<sub>Uploaded with [Argos ↗︎](https://argos-ci.com/docs/learn/media/standalone-media-upload). This comment is updated in place.</sub>",
+  ].join("\n");
+}
+
+/** A pipe inside a cell would split it into two columns. */
+function escapeCell(value: string): string {
+  return value.replace(/\|/g, "\\|");
```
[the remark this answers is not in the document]
gregberge: Fixed in b882cffd — this file's table rendering moved into `apps/backend/src/media/url.ts`, where the same helper had the same hole. Details in the reply on alert 120.

apps/backend/src/api/handlers/media.ts
gregberge: Reverted to your call in b441ec02 — `GET /media/{mediaId}/versions`, and `versions` is off the media payload.

You were right about the trade. Inlining paid for the rare case on every read: most media have one version, and most callers want the newest, which the media already carries flattened onto it.

`versionCount` comes back on the media as the signal — a caller seeing 1 knows there is nothing to fetch. And the half that made this worth solving stays: a comment carries `mediaId` and `mediaVersionId`, which it did not before, so the endpoint has something to resolve against.

apps/backend/src/api/handlers/media.ts
gregberge: Renamed to `staged` in b441ec02 — values, prose and test names. Agreed it is clearer: `draft` suggests unfinished, and a staged media is not. It is real, uploaded and shareable; the only thing missing is the pull request.

`github_pull_requests.draft` is GitHub's own flag on a pull request and is untouched.

jsfez (approved):

PR #2458, opened 2026-08-11
feat(media): compare a before/after pair and mark what changed

gregberge:
Closes ARG-490.

Runs the same diff engine a build uses (odiff) over the two halves of a before/after media pair, in a background job, and draws the resulting mask on the share page with the build's own overlay controls.

## The data model is the point

A `media_diffs` row is keyed on the two **versions** it was computed from, not on the two media. That is what makes it impossible to show a stale answer: re-uploading either half is a new pair, so it is a new row, and until that one is computed the page shows *no* overlay rather than the previous one. Nothing is ever recomputed in place, which also makes the job safely re-runnable and means the mask a reviewer is looking at cannot change under them while a job runs.

`key` is the mask — a PNG whose opaque pixels are the ones that differ, exactly like `screenshot_diffs.s3Id`. It is null when the halves are identical (`score` 0) and while the job has not run, which `jobStatus` tells apart. Deliberately not a `files` row: media never joins that table, and the machinery around it (fingerprints, ignored changes, test stats) is build-only.

## Backend

- `media/diff.ts` — the comparison itself. Content-addressed mask under `media/<projectId>/diffs/<sha256>.png`, so two pairs that changed the same way share one object. Identical halves are answered from their keys without downloading anything.
- `media/diff-job.ts` — the queue worker, same 60s budget a screenshot diff gets.
- `media/diff-schedule.ts` — the idempotent insert-then-queue. Atomic rather than check-then-insert, so two uploads landing at once do not both queue; only the caller whose insert actually produced a row pushes the job.
- `media/pair.ts` — the pairing tuple, extracted so the GraphQL loader's batched version and the scheduler's single lookup cannot drift apart.

Scheduling fires **where a half is renewed** (when an upload's bytes land) and **again when a share page is opened**, so a pair that came together some other way — a half adopted onto a pull request, a pair that predates this change — is picked up too. Both paths are best-effort: a queue that is down never fails an upload that has already succeeded, and the row is taken back rather than left `pending` forever, so the next attempt starts clean.

Retention purges the masks along with the versions they were computed from, keeping any a surviving pair still shows (mask keys are content-addressed too).

## Frontend

The controls are **shared, not reimplemented**. `ChangesOverlayControls`, `ChangesMask` and `ChangesHighlights` are extracted out of the build's diff viewer into `containers/Build/ChangesOverlay.tsx` and used by both surfaces, so show/hide, the colour and opacity settings, the highlight and next/previous-change buttons and their `D` / `H` / `J` / `K` shortcuts behave identically on a build and on a share page. Works in all four view modes — single, side by side, onion and swipe.

While the comparison is queued the page polls for it (capped at two minutes) and the overlay appears on its own; switching the version picker off the newest version withdraws it, because the mask was computed from other bytes.

## Two alignment fixes fall out of this

odiff pads both halves to a common canvas from the **top left** before comparing, so the mask shares the images' pixel grid on a possibly taller canvas. Both the mask and — this is the fix — the blended counterpart layer are now placed on the base image's grid. That layer used to be `object-contain`-fitted and centred, so with a mask on top every pixel it marked sat over a different pixel whenever the two halves differed in height, which for full-page screenshots is the common case rather than the corner one.

## Notes for the reviewer

- The seeded pair was byte-identical, so it compared to nothing and the whole surface had no changes to show. Its halves now genuinely differ, which is why `media.spec.ts` expects `375×1024` where it expected `375×720`.
- The REST/OpenAPI surface does not expose the diff — the issue scoped this to the UI. Easy to add if the CLI should see it.

## Verification

`pnpm run static-checks` clean. 1167 integration tests, 600 unit tests and 91 Playwright tests pass. New coverage: 10 scheduling cases (idempotence, renewal, branch isolation, video refusal, queue failure) and 2 compute cases against real storage, which assert the mask object actually lands rather than trusting the row that names it.


apps/backend/src/api/handlers/media.ts
```diff
@@ -422,9 +426,18 @@ export const deleteMediaHandler: CreateAPIHandler = ({ delete: del }) => {
...
+    const diffs = await getMediaDiffObjects(versionIds);
     await deleteUnreferencedMediaObjects({
       keys: versions.map((version) => version.key),
-      excludeVersionIds: versions.map((version) => version.id),
+      excludeVersionIds: versionIds,
+    });
+    await deleteUnreferencedMediaDiffObjects({
+      keys: diffs.keys,
+      excludeDiffIds: diffs.diffIds,
     });
```
gregberge: Is it possible to run this in parallel?

apps/backend/src/api/handlers/media.ts
```diff
@@ -422,9 +426,18 @@ export const deleteMediaHandler: CreateAPIHandler = ({ delete: del }) => {
...
     await media.$query().delete();
```
gregberge: I think the whole thing should run in a transaction?

apps/backend/src/graphql/services/project.ts
```diff
@@ -232,4 +245,8 @@ export async function unsafe_deleteProject(args: {
     keys: mediaKeys,
     excludeVersionIds: [],
   });
+  await deleteUnreferencedMediaDiffObjects({
+    keys: mediaDiffKeys,
+    excludeDiffIds: [],
+  });
```
gregberge: What about the transaction here?

PR #2472, opened 2026-08-13
chore(agents): centralize agent identity in one package

gregberge:
Stacked on [#2471](https://github.com/argos-ci/argos/pull/2471) — review that one first; this diff is the refactor alone.

The same handful of coding agents were described in **three** places, each with its own ids, labels and SVGs, and nothing kept them in step:

| Where | What it knew | Ids |
| --- | --- | --- |
| `oauth/known-apps.ts` | verified badge, official name, homepage | `claude`, `claude-code`, `openai-codex`, … |
| `agent/registry.ts` | comment attribution, display names, aliases | the same, minus `argos-cli`, plus 8 more |
| `util/ai-agents.ts` | prompt deep links, labels, icons | `claude-desktop`, `codex`, `cursor` |

Plus two icon sets — `oauth-logos.tsx` (brand colors) and `agent-icons.tsx` (`currentColor`) — with three brands drawn in both.

`@argos/agents` is one entry per product: id, name, homepage, the names it reports for itself, the OAuth signals it is recognized by, its prompt deep link — and both sets of marks behind `AgentLogo` / `AgentIcon`.

## Perf: two entry points that share nothing

The backend resolves an agent on **every comment it stores**, so the data must not drag React or a few kB of SVG in with it.

| Import | Contains | Built size | Used by |
| --- | --- | --- | --- |
| `@argos/agents` | registry + resolvers, no React | 6.7 kB | backend, frontend |
| `@argos/agents/react` | the marks and the two components | 18.6 kB | frontend only |

`dist/index.js` and `dist/react.js` emit **no shared chunk** — that is why `AgentLogo` takes the `name` for its monogram as a prop instead of looking it up. React, `lucide-react` and `clsx` are peer deps, so the app's copies are used and nothing duplicates into the bundle.

## Two things a reviewer should check

**Tailwind.** Auto-detection stops at `node_modules`, and workspace packages resolve through it, so classes on the marks would have generated no CSS. `index.css` gains one `@source` pointing at the package source. The consent screen in the Playwright run confirms it — the Argos CLI mark renders with its border and background.

**A stored preference had to move.** `aiPromptTarget` in localStorage held `claude-desktop` / `codex`, which are now `claude` / `openai-codex`. Left alone, anyone who had picked one would silently drop back to "copy" with nothing to correct it, so `useAiPromptTarget` maps the two legacy values over.

Ids are otherwise untouched — they are persisted in `comments.agent` and `oauth_clients.knownAppId`, so the package documents them as append-only.

## Behavior

Preserved deliberately, including the quirk where the OAuth match order decides between two apps sharing a redirect host: the entries keep their relative order and `resolveOAuthAgent` still returns the first match.

One copy change, from folding two names into one: the consent screen now says **"Codex"** rather than "OpenAI Codex" (the name the prompt menu already used). Say the word if you'd rather it went the other way — it's one field.

## Verified

`pnpm run static-checks`, unit (690) and integration (1189) suites green. `packages/agents/src/registry.test.ts` covers all three doors — reported names with their version tags, OAuth matching and what must *not* match, the prompt list — and takes over the known-app cases that were in `oauth.test.ts`.

Playwright: the agent badge, the prompt menu and the OAuth consent screen all pass, and the badge screenshot is pixel-identical to #2471's.

Flakes seen once each under full-suite load and passing on re-run and in isolation: `testChanges`, the MCP scope test, the image-diff timeout, the server card. Pre-existing, unrelated to this diff.


gregberge:
Its' not urgent, let's not do it.

PR #2493, opened 2026-08-17
feat(ui): move the overlays onto Base UI

gregberge:
Batch 1 of the remaining react-aria → Base UI migration: **Modal, Dialog, Popover, EmojiPicker** — the batch everything else waits on.

## The three things that were not renames

**Pending blocks dismissal.** `Modal` monkey-patched `OverlayTriggerStateContext` to intercept close requests, because react-aria owned the open state. Base UI hands it to us, so `ModalCloseTracker` disappears entirely: a *user* dismissal (Escape, backdrop, trigger) arrives in `onOpenChange` and is refused with `details.cancel()` while an action runs, while the app's own `close()` drives the state directly and stays allowed. `ModalActionContext` keeps its exact shape, so `Form`, `DialogActionButton` and `DialogDismiss` needed nothing. (There is no `disableEscapeDismissal` prop — Escape arrives like any other dismissal, so one branch covers all three.)

**The overlay surface splits three ways** — tint/blur/z-index onto `Dialog.Backdrop`, centering onto `Dialog.Viewport`, the card onto `Dialog.Popup`. `h-dvh` replaces the `--visual-viewport-height` measurement react-aria ran a script for.

**The dialog labels itself.** `Heading slot="title"` is gone, so `DialogTitle` takes an id from the dialog and `Dialog` points `aria-labelledby` at it. The `dialog`/`alertdialog` role stays on `Dialog` with the popup around it as `presentation`, so `getByRole("dialog")` still finds exactly one element — the four element-scoped dialog screenshots in `passkeys.spec.ts` keep working unchanged.

## Deliberate behaviour changes

- **The exit fades as it shrinks.** react-aria tore the card out on the animation's last frame, so the missing fade never showed; Base UI keeps it mounted a beat longer and a card that only scales visibly pops out at the end.
- **The entrance grows from 95%** instead of shrinking from 105%. It deliberately does **not** fade: the card must be opaque on every frame it is on screen, or the page shows through it — which is exactly what a Firefox screenshot caught mid-entrance when I tried it. The backdrop does the fading, as it always did.
- **Focus lands on the dialog itself**, as it did on react-aria. Base UI's default focuses the first tabbable element, which would put a focus ring on *Cancel* in every dialog screenshot. Verified it does not clobber `autoFocus`ed inputs — the focus manager bails when focus already moved inside.

## `ui/Overlay.tsx` — why a local file exists

Base UI's dialog namespace exports only parts plus `createHandle`; there is **no hook or context for a descendant to close the overlay it lives in**. 48 sites need exactly that, nearly all of the shape `onCompleted: () => state.close()` — a mutation callback several components below the `Modal`. `Dialog.Close` is a button and cannot express it; `actionsRef` belongs to whoever renders the Root; a handle is module-scoped. So the close context is a genuine gap-filler, not a compatibility shim.

What *is* inherited from react-aria is the **shape**: `<DialogTrigger>` taking trigger and overlay as siblings, and the `useOverlayTriggerState()` name. That keeps 28 files unchanged, at the cost of one wart — the trigger element is held in context and rendered by the overlay, which means `getTriggerRenderer` special-cases `Tooltip`/`HotkeyTooltip` by component identity (3 real call sites wrap their trigger in one). Base UI's native composition — a `<ModalTrigger>` part inside `<Modal>` — would delete that wart, for ~28 files of churn. Worth doing with the Button batch; flagging rather than deciding unilaterally.

## Two silent breaks found in review

`slot="close"` and `slot="title"` were react-aria mechanisms that compile fine and do nothing now. The hotkeys dialog's close button was dead (fixed in `chore: fix hotkey dialog`), and the overlay settings dialog was left with **no accessible name** — `aria-labelledby` pointing at an id nothing rendered. Both fixed, and a dialog with neither a `DialogTitle` nor an `aria-label` now warns in development, because nothing else can see it: not a screenshot, not a type, and a dangling reference reads as a name right up until a screen reader asks.

## What's left behind on purpose

`ui/Popover` keeps a react-aria `SelectPopover`: `Select` reads its state from react-aria context, so it cannot move until `ListBox` does (batch 2). Its 16 call sites are renamed to say so, which also gives the next batch an exact worklist.

`react-stately` leaves the lockfile — the first of the four packages to go.

## Verification

`static-checks` 11/11 · Storybook **76/76** · full chromium e2e **96 passed** (the 2 local reds were the backend rate-limiter serving "Too many requests"; both pass in isolation).

**Argos, reviewed:** Storybook **1 of 76** changed — `ui-popover--open`, where only the `right`-placement popover shifts ~2px; every dialog snapshot is pixel-identical. e2e **2 of 239** changed, and the 94% one on `firefox/settings/passkeys-create` is what caught the entrance-fade problem above; it should be back to identical on this run.

The pending-dismissal contract is the one behaviour no screenshot can see, so it now has a play test (`Dialog > PendingBlocksDismissal`) that starts an action, asserts Escape / backdrop / dismiss-button are all refused, settles the action, and asserts the dialog un-blocks.

### DIFF
--- a/apps/frontend/src/containers/Build/toolbar/SettingsButton.tsx
+++ b/apps/frontend/src/containers/Build/toolbar/SettingsButton.tsx
@@ -37,9 +37,11 @@ export const SettingsButton = memo(() => {
 
 function OverlaySettingsDialog() {
   return (
-    <Dialog className="w-80 select-none">
+    <Dialog className="w-80 select-none" aria-label="Customize overlay">
       <DialogBody>
-        <Heading slot="title" level={2} className="mb-4 font-medium">
+        {/* Not a `DialogTitle`: this one reads at body size. The dialog is
+            named by `aria-label` instead, which says the same words. */}
+        <Heading level={2} className="mb-4 font-medium">
           Customize overlay
         </Heading>
         <div className="flex flex-col gap-6">
--- a/apps/frontend/src/containers/Build/toolbar/SettingsButton.tsx
+++ b/apps/frontend/src/containers/Build/toolbar/SettingsButton.tsx
@@ -2,11 +2,10 @@ import { memo } from "react";
 import { invariant } from "@argos/util/invariant";
 import { useAtom } from "jotai/react";
 import { PaintbrushIcon } from "lucide-react";
-import { Heading } from "react-aria-components";
 
 import { Button } from "@/ui/Button";
 import { ColorSwatchPicker, ColorSwatchPickerItem } from "@/ui/ColorPicker";
-import { Dialog, DialogBody, DialogTrigger } from "@/ui/Dialog";
+import { Dialog, DialogBody, DialogTitle, DialogTrigger } from "@/ui/Dialog";
 import { Label } from "@/ui/Label";
 import { Popover } from "@/ui/Popover";
 import {
@@ -37,13 +36,13 @@ export const SettingsButton = memo(() => {
 
 function OverlaySettingsDialog() {
   return (
-    <Dialog className="w-80 select-none" aria-label="Customize overlay">
+    <Dialog className="w-80 select-none">
       <DialogBody>
-        {/* Not a `DialogTitle`: this one reads at body size. The dialog is
-            named by `aria-label` instead, which says the same words. */}
-        <Heading level={2} className="mb-4 font-medium">
+        {/* Reads at body size rather than a dialog title's, but still names
+            the dialog. */}
+        <DialogTitle render={<h2 className="mb-4 font-medium" />}>
           Customize overlay
-        </Heading>
+        </DialogTitle>
         <div className="flex flex-col gap-6">
           <ColorPicker />
           <OpacityPicker />
--- a/apps/frontend/src/containers/Build/toolbar/SettingsButton.tsx
+++ b/apps/frontend/src/containers/Build/toolbar/SettingsButton.tsx
@@ -28,7 +28,7 @@ export const SettingsButton = memo(() => {
           <PaintbrushIcon />
         </Button>
       </Tooltip>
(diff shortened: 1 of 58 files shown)


apps/frontend/src/containers/Build/toolbar/SettingsButton.tsx
```diff
@@ -37,9 +37,11 @@ export const SettingsButton = memo(() => {
...
   return (
-    <Dialog className="w-80 select-none">
+    <Dialog className="w-80 select-none" aria-label="Customize overlay">
       <DialogBody>
-        <Heading slot="title" level={2} className="mb-4 font-medium">
+        {/* Not a `DialogTitle`: this one reads at body size. The dialog is
+            named by `aria-label` instead, which says the same words. */}
+        <Heading level={2} className="mb-4 font-medium">
           Customize overlay
         </Heading>
```
gregberge: In this case we should expose a headless DialogTitle to style it as we want. It does not seem the good way of migrating things.

gregberge: Agreed — the `aria-label` was the wrong shape: it duplicated the words the heading already said, and left the visible heading naming nothing.

`DialogTitle` now takes a `render` element and hands it the id that names the dialog, which is the same escape hatch Base UI's own parts offer:

```tsx
<DialogTitle render={<h2 className="mb-4 font-medium" />}>
  Customize overlay
</DialogTitle>
```

Default callers are untouched. This also drops the last react-aria `Heading` import from that file.

Pinned with a play test rather than left to trust, since a dangling `aria-labelledby` reads as a name right until a screen reader asks: it asserts the dialog is findable by its styled title, and that the caller's styling won over the default `text-xl`.

PR #2511, opened 2026-08-19
feat(flows): a Flows tab — every E2E test, and the screens it captures

jsfez:
Argos only ever heard about a test that **took a screenshot**. The screenshot arrived carrying its test in `metadata`, so a test that captured nothing was indistinguishable from a test that did not exist — which means "this test has no visual coverage" was never a sentence the product could say. There was no denominator.

The test runner's report has it. This PR stores it and builds two surfaces on top.

## What lands in the database

`metadata.testReport.tests` now carries **every test of the run** — the ones that captured nothing and the ones that were skipped included — and it becomes two tables:

- `flows` — the identity of a test across builds (its title path, starting at the file).
- `flow_runs` — one execution of it in one build.

The split is the one `tests` / `screenshots` already makes, for the same reason: the identity outlives the run, and only the run has a status. It is what lets a failure rate be read across builds through a foreign key rather than by matching title strings on every query.

A flow is deliberately blind to the runner project: one test walked under chromium, firefox and webkit is **one test, run three times**. The browser lives on the run, mirroring how `variantKey` already collapses the variants of a screenshot.

The report is consumed into rows and **stripped from the persisted metadata** — keeping it would store a second, unqueryable copy of a payload that reaches tens of thousands of entries, and the shard aggregation would copy it again.

Screenshots are attached to their run at **finalization**, through the runner's own test id (`metadata.test.id`, which the report repeats). Nothing orders the report against the screenshots — a parallel build can upload a shard's screenshots before the shard that reports them — so any earlier moment would miss half of them.

## The Flows tab

Every test the reference build ran, grouped by spec file, with the failure and flaky rates of the last 30 days.

There is deliberately **no coverage percentage**. A test that legitimately captures nothing — a redirect guard, an API check — would drag a score down for being right, and a metric whose correct value nobody knows directs no action. A test with no screenshot simply leaves its cell empty, so the gap is what stands out rather than a column of zeros.

Rows carry the `describe` blocks a test is nested in: a suite that nests them ends up with rows reading "works", and two tests named the same in two describes would otherwise be the same row twice.

## The flow page

**A flow is a spec file** — every test in it, and the screens they captured, on one canvas.

It was first keyed on a folder shared by the screenshot names (`screenshots(page, "supplier-invoice")` giving `supplier-invoice/…`). That is a convention a suite may or may not follow, and most do not: their names are flat, so every test would be a flow of one and the diagram would never show a path. Every test has a file.

The canvas is `@xyflow/react` (new dependency, lazy-loaded):

- **Scrolling pans, pinching zooms.** The wheel deliberately does not zoom, which would make the canvas jump every time the page scrolls under it.
- **Names keep their on-screen size** whatever the zoom, and are anchored above what they name so growing one moves it away from the screenshot rather than over it. A screenshot of a long page is very tall, which is also why the name sits above the image.
- **One violet frame per test**, so a path walked by three tests reads as three segments of one line.

## Variants

`viewports: ["iphone-xr", "macbook-13"]` turns one `argosScreenshot` call into two files. The canvas folds them through `getVariantKey`, the same way Argos folds them everywhere else: **one node per screen**, and a select per dimension that actually varies (viewport, browser, theme) to choose which capture it shows. The choice is in the URL, so the link opens on what the sender was looking at.

A filter rather than lanes side by side: mobile against desktop is not a regression question — they are two intended designs — and lanes multiply, since two viewports, a dark mode and three browsers is twelve rows of the same path.

A screen the chosen variant has no capture for **keeps its place**, marked with what it is missing. That is a viewport coverage gap, and the tab exists to make gaps visible.

The runner project prefix is stripped by the name the run reports, not by `getVariantKey`'s hard-coded list of browsers — a Playwright project is free to be called `Desktop Chrome`. `variant-key.ts` moves to `util/`, where a pure string function belongs and where both layers may import it.

## Note on `capture.trigger`

Added to the screenshot metadata schema so the API accepts it before the SDK sends it. Without it, a test that fails under `screenshot: "only-on-failure"` counts as captured: Playwright takes a screenshot on its own, the reporter uploads it with the test's metadata, and only the ` (failed)` name suffix tells it apart. The SDK side ships separately.

## Scope

Deliberately out:

- **The SDK** — the reporter does not send `testReport.tests` yet. The field is optional, so nothing breaks; the tab is empty until it does, and the seed covers local development and the specs.
- **Visual flakiness as a column** — Argos already computes it, but on screenshots, through a separate metrics subsystem with its own period model. Worth showing next to the test-level one (a stable test whose screenshots flicker is the opposite problem from a flaky test whose screenshots are stable). Reachable without new columns: flow → runs → screenshots → `tests`.
- **The PR check** ("this PR adds 2 tests that capture nothing").
- **Removed tests** — `flows.lastSeenAt` is written for it, nothing reads it yet. A test that stops running disappears from the list silently today, which is the coverage loss most worth surfacing next.

## Tests

- `tests/project-flows.spec.ts` — the list including the tests that capture nothing, the describe path, the screenshots filter, a flow across three tests, a flow of one, a file where nothing captures, and the variant select with the gap it leaves.
- Unit tests on the two pure rules: `getStepKey` (runner prefix, viewport, theme, repeat suffixes) and `getScreenVariant`.
- Full integration suite green (1220 tests), and `pnpm run static-checks` clean.
- Seed: `createFlowsScenario` on `smooth/big`, modelled on a real loan-request suite — a flow across three tests, a flow of one, tests that capture nothing, a skipped test with its reason, and one screen captured on desktop only.

## Unrelated finding

`structure.sql` and the migrations have diverged on `github_pull_requests.id`: the committed dump says `bigint`, a fresh `migrate:latest` produces `integer`. The app depends on `bigint` (Objection expects the id as a string, which is how `pg` returns `int8` and not `int4`), so a schema regenerated from the migrations breaks `createBuild`. Left alone here — this PR's dump is built on top of the committed one — but it is worth fixing on its own.


apps/backend/db/migrations/20260819140722_flows-test-coverage.js
```diff
@@ -0,0 +1,124 @@
...
+ * A *flow* deliberately ignores the test runner project (chromium, firefox, a
+ * device): the same test walked under three browsers is one journey seen three
+ * times, not three journeys. The browser lives on the run, which mirrors how
+ * `variantKey` already collapses the browser and viewport variants of a
+ * screenshot.
+ *
+ * @param {import('knex').Knex} knex
+ */
+export const up = async (knex) => {
+  await knex.schema.createTable("flows", (table) => {
```
gregberge: I think we have to store the exact structure of Playwright here not a generic one.

gregberge:
I don't want to go into flow like this for now. I prefer to integrate test reports from Playwright and to build flows on top of it later.

PR #2514, opened 2026-08-21
fix(build): stop naming the build in the next-review button

gregberge:
Closes ARG-505.

## The problem

The next-review prompt named its call to action after the build it points at — `Review {build.name}`. That word is `default` whenever CI never named the suite, since `default` is what Argos stamps on an unnamed one. A reviewer coming out of one project's build was offered **Review default**, which says nothing about where it goes.

## The fix

The button no longer names anything: **Review next build**.

Nothing takes the name's place, deliberately. The rows above are where a build is picked out, and they have two lines each to do it with — the build name, and the project or the number underneath. Naming the first of them on the button only raises the question of what the word means: `default` says nothing, and anything better (the project it ran in, its number) is a rule the reader has to infer from a single word.

| Before | After |
| --- | --- |
| [before](https://app.argos-ci.com/m/951ijtbhyq2m65dcyw6k) | [after](https://app.argos-ci.com/m/pzlk24geybxstlvyp9yj) |

The reviewer is on `acme/sparkle-docs`, and the next build in line is `acme/sparkle`'s unnamed suite.

## Tests

The two existing next-build specs now assert the static label, and one of them clicks it and follows the navigation, so the button is still covered end to end. No new test: there is no longer a rule to guard.

### DIFF
--- a/apps/frontend/src/pages/Build/BuildNextReviewDialog.tsx
+++ b/apps/frontend/src/pages/Build/BuildNextReviewDialog.tsx
@@ -197,7 +197,13 @@ function BuildNextReviewDialog(props: { build: ReviewedBuild }) {
           href={getBuildOverviewURL(getBuildParams(nextBuild))}
           autoFocus
         >
-          Review {nextBuild.name}
+          {/*
+           * Unnamed: the rows above are where a build is picked out, and they
+           * have two lines to do it with. Naming the first one here only
+           * raises the question of what the word means — `default` says
+           * nothing, and anything better takes a rule the reader has to infer.
+           */}
+          Review next build
         </LinkButton>
       </DialogFooter>
     </Dialog>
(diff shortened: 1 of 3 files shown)


jsfez (approved):

apps/frontend/src/pages/Build/BuildNextReviewDialog.tsx
```diff
@@ -197,7 +197,13 @@ function BuildNextReviewDialog(props: { build: ReviewedBuild }) {
           href={getBuildOverviewURL(getBuildParams(nextBuild))}
           autoFocus
         >
-          Review {nextBuild.name}
+          {/*
```
gregberge: Remove this comment and in general, add something into AGENTS.md to make sure to only comment the revelant parts in the code. This comment is useless here. It does not help to understand the code.

jsfez (approved):

PR #2525, opened 2026-08-22
feat(build): undo and redo review marks with ⌘Z / ⌘⇧Z

jsfez:
Marking a snapshot takes the reviewer straight to the next one, so a mis-hit on <kbd>Y</kbd> or <kbd>N</kbd> is always noticed one snapshot too late — and the only way back was to navigate there by hand and toggle the mark off.

<kbd>⌘Z</kbd> / <kbd>Ctrl+Z</kbd> walks back the last marking and brings the snapshot it belongs to back on screen. <kbd>⌘⇧Z</kbd> / <kbd>Ctrl+Shift+Z</kbd> walks forward again. Both are listed in the `?` dialog under **Actions**, and each step confirms itself with a toast naming the state the changes are in now ("1 change is pending review again").

## Is ⌘Z free?

Yes. The build page only takes `⌘F`, and the media page `⌘C`, `⌘⇧,` and `⌘⇧D`. Nothing claims `⌘Z` or `⌘⇧Z`, and the browser has no page-level undo to compete with. The document listener already skips text inputs and anything under `[data-hotkeys-disabled]`, which is what the comment editor sets — so `⌘Z` inside a comment still undoes typing, as it should.

## What is on the stack, and what is not

The stack is **the reviewer's local, not-yet-submitted marks** — the state they build up before hitting submit. That is one coherent thing, and the only thing in the flow that is both cheap to change by accident and free to change back.

Recorded:

- accepting or rejecting a single snapshot;
- accepting or rejecting a collapsed group, which marks every snapshot in it;
- **reapplying previous approvals** — the one action that marks dozens at once from a single click, and by far the one most worth being able to take back wholesale.

Deliberately left out, because they already reached the server:

- **Submitting the review.** It clears the stacks instead: undoing into a pre-review state would only misrepresent what was submitted.
- **Ignoring a change.** Ignoring also marks the snapshot accepted, and *that mark* is recorded like any other — but undo does not un-ignore. An ignore is a project-wide rule, confirmed through its own dialog, with its own toggle to reverse it; pulling it into a keystroke-level stack would make `⌘Z` mean two very different things.
- **The note explaining a rejection.** Rejecting invites a note, which is a build comment the moment it is submitted. Undo takes the mark back but leaves the note, so the toast says so — otherwise the snapshot would quietly keep a written justification for a rejection that no longer exists.
- **Comments, navigation, view toggles.** Comments have their own controls; navigation is what the browser's back button is for; view state is not a change to anything.

## Why redo

Undo without redo is a trap: overshoot by one <kbd>⌘Z</kbd> and the mark can only be restored by hand, which is the problem this is meant to solve. It costs one more entry on a stack that already exists. `⌘⇧Z` is the one combination that works on every platform, so no separate `Ctrl+Y` — every browser and every design tool binds redo the same way.

## Design notes

- **The review state owns marking.** `markDiffs` / `undo` / `redo` / `clearHistory` live on `BuildReviewAPI`, with the two stacks as an internal detail of the provider. `markDiffs` writes the statuses *and* records the step, so marking a diff off the record is not something a future change can do by accident.
- The stacks are **in memory and scoped to one build**. The marks themselves survive a reload (they are stored per build), but a list of steps to walk back does not belong to a page the reviewer has left. The provider stays mounted across builds, so the stacks carry the build they were recorded on and reset when it changes.
- Steps store only the diffs they touched, so undoing never resurrects marks they left alone.
- Undo and redo both **navigate to the snapshot involved**. A mark you cannot see change is a mark you cannot trust.
- The stacks live in a ref: nothing renders from them, and recording a step must not re-render the page mid-review. Capped at 50 steps.

## Two fixes the feature dragged in

**Shift was demanded but never forbidden** by the matcher, so `⌘⇧Z` matched `⌘Z` too and every redo also ran an undo. The two happened to cancel out on state, but produced two toasts and a jump to the wrong snapshot.

Shift is now forbidden unless declared, with one exception: a hotkey named by the character it prints already carries shift in `event.key`, so demanding it twice would make it impossible to type. That covers `?` (shift+/) and the **digits**, whose row is shifted on an AZERTY keyboard — pressing the `1` the dialog names sends `Digit1` with shift held. The digits are declared as `"1"` rather than `"Digit1"` for exactly that reason, which also retired the `hasDigits` alt special-case. The tightening stops `⌘⇧F` from opening "Find snapshot" and `⇧Y` from accepting.

**The dialog told Windows and Linux readers to press ⌘.** The shortcut always worked there, but only because the matcher converted `"⌘"` to `ctrlKey` off macOS while `displayKeys` spelled the label with Mac symbols in plain strings — two representations of one key, one of them wrong.

`MOD` / `ALT` / `SHIFT` (moved next to `isMacOS` in `util/os`, out of `ui/Editor/EditorToolbar.shortcuts` where three non-editor files were already reaching for them) now name the modifier for both halves: `keys: [MOD, "KeyZ"]` and `displayKeys: [MOD, "Z"]`. A modifier-name → event-flag lookup replaces the conversion, so nothing below `util/os` knows which platform it runs on — `isMacOS` has no callers left outside it. `⌘F`, `⌘C`, `⌘⇧,` and `⌘⇧D` were wrong the same way and are fixed with it.

| macOS | Windows / Linux |
| --- | --- |
| Undo last review mark · <kbd>⌘</kbd> <kbd>Z</kbd> | Undo last review mark · <kbd>Ctrl</kbd> <kbd>Z</kbd> |
| Redo last undone review mark · <kbd>⌘</kbd> <kbd>⇧</kbd> <kbd>Z</kbd> | Redo last undone review mark · <kbd>Ctrl</kbd> <kbd>Shift</kbd> <kbd>Z</kbd> |

## Testing

`containers/Build/hotkeys.test.ts` — a table of (platform, hotkey, event) → matches over the real declarations: ⌘Z vs Ctrl+Z per platform, ⌘Z vs ⌘⇧Z, the shifted AZERTY digit, and the labels each platform shows. 15 cases in 135ms. The declarations and the matcher moved to `containers/Build/hotkeys.ts` so the test does not pull React in.

`tests/build.spec.ts` keeps the one spec that earns an end-to-end: marking two snapshots, walking both back and both forward, checking each step lands on the right snapshot with the right mark. It fails without the shift fix — the last redo lands on the wrong snapshot.

Every guard was verified to fail with its fix reverted. `build.spec.ts` (43) and `media.spec.ts` (25) pass on chromium and firefox, as does `pnpm run static-checks`.

## Known gap

Undoing while a *reapply previous approvals* acknowledgment is still held — the build is large and its diffs are still streaming — can release that acknowledgment and race its navigation against the undo's own, landing the reviewer on the wrong snapshot. The window is narrow and nothing is lost but the scroll position. Fixing it properly means changing how `useAcknowledgeMarkedDiff` detects that its own change landed, which is a bigger change than this PR should carry.

## Open question

No visible affordance beyond the `?` dialog and the toasts. The snapshot toolbar is already carrying ten controls, and I did not want to add two more for something that is a keyboard reflex — but if you would rather have undo/redo buttons somewhere, the review sidebar is probably the place.


apps/frontend/src/util/os.ts
```diff
@@ -1,3 +1,17 @@
...
+ *
+ * macOS names its modifiers with symbols and nothing else; every other
+ * platform spells them out, and a Windows keyboard has no ⌘ key at all. Only
+ * the label differs — `MOD` is Command or Control depending on where the app
+ * is being read, which is the same key in both cases as far as the shortcut
+ * is concerned.
+ */
+export const MOD = isMacOS ? "⌘" : "Ctrl";
+export const ALT = isMacOS ? "⌥" : "Alt";
+export const SHIFT = isMacOS ? "⇧" : "Shift";
```
gregberge: I think it's a very good thing but in this case we should use them everywhere. Currently it's used in `displayKeys` but not in `keys` for instance.

So we should remove the logic that converts `⌘` to `Ctrl` in the shortcut runner and use these, no more magic.

apps/frontend/src/containers/Build/BuildHotkeys.tsx
```diff
@@ -325,6 +342,22 @@ function checkIsModifiedPressed(event: KeyboardEvent) {
...
+/**
+ * Whether shift counts as pressed for `hotkey`.
+ *
+ * Shift is only meaningful for hotkeys written as physical codes (`KeyZ`,
+ * `Digit1`, `ArrowUp`…), where the declared `⇧` is the only thing that can
+ * demand it. Hotkeys written as literal characters ("?", "[") already carry
+ * the answer in `event.key` — "?" *is* shift+/ — so demanding a matching
+ * `shiftKey` would make them impossible to type.
+ */
+function checkIsShiftPressed(hotkey: Hotkey, event: KeyboardEvent): boolean {
```
gregberge: I am not sure to understand why we need it now and not before.

apps/frontend/src/pages/Build/BuildPreviousReviewDialog.tsx
```diff
@@ -104,16 +106,20 @@ function ReapplyPreviousApprovalsButton(props: {
...
+        for (const diffId of branchApprovedDiffs) {
+          before[diffId] = currentStatuses[diffId] ?? EvaluationStatus.Pending;
+          after[diffId] = EvaluationStatus.Accepted;
+        }
+        api.setDiffStatuses((prev) => ({ ...prev, ...after }));
+        // Reapplying is the one action that marks dozens of snapshots at once,
+        // which is exactly the one worth being able to take back wholesale.
+        const [firstDiffId] = branchApprovedDiffs;
+        if (firstDiffId) {
+          history?.record({ before, after, diffId: firstDiffId });
```
gregberge: Why using a conditional operator here, in which case shouldn't we have a history?

apps/frontend/src/containers/Build/BuildHotkeys.tsx
```diff
@@ -325,6 +342,34 @@ function checkIsModifiedPressed(event: KeyboardEvent) {
...
+  if (key === "⌘" || key === "⌥" || key === "⇧") {
+    return false;
+  }
+  return key.length === 1 || key.startsWith("Digit");
+}
+
+/**
+ * Whether `event`'s shift state matches what `hotkey` asks for.
+ */
+function checkShiftMatches(hotkey: Hotkey, event: KeyboardEvent): boolean {
```
gregberge: Why do we suddenly need all this logic?

apps/frontend/src/pages/Build/BuildPage.tsx
```diff
@@ -105,6 +106,10 @@ export const BuildPage = ({ params }: { params: BuildParams }) => {
             <BuildNextReviewDialogProvider>
               <BuildReviewDialogProvider project={data?.project ?? null}>
                 <RejectCommentDialogProvider build={build}>
+                  {/* Mounted here rather than beside the review buttons: the
+                      keys have to answer from the overview too, where marks are
+                      reapplied in bulk and no snapshot toolbar exists. */}
```
gregberge: Remove this comment.

apps/frontend/src/pages/Build/BuildPreviousReviewDialog.tsx
```diff
@@ -103,17 +104,8 @@ function ReapplyPreviousApprovalsButton(props: {
...
-          ...branchApprovedDiffs.reduce<Record<string, EvaluationStatus>>(
-            (acc, diffId) => {
-              acc[diffId] = EvaluationStatus.Accepted;
-              return acc;
-            },
-            {},
-          ),
-        }));
+        invariant(history);
+        history.markDiffs(branchApprovedDiffs, EvaluationStatus.Accepted);
```
gregberge: It's not the responsibility of the history to mark diff statuses.

apps/frontend/src/pages/Build/BuildReviewUndoHotkeys.tsx
```diff
@@ -0,0 +1,98 @@
...
+      return;
+    }
+    const diff = diffs.find((diff) => diff.id === change.diffId);
+    if (diff) {
+      setActiveDiff(diff, true);
+    }
+    toast.success(describeAppliedChange(change), {
+      description: checkTakesBackRejection(change)
+        ? "Any note you left explaining the rejection stays on the snapshot."
+        : undefined,
```
gregberge, line 90: Good

apps/frontend/src/pages/Build/BuildReviewHistoryHotkeys.tsx
```diff
@@ -0,0 +1,98 @@
...
+    });
+  });
+
+  const enabled = Boolean(history);
+  useBuildHotkey("undoReviewMark", () => apply(history?.undo() ?? null), {
+    enabled,
+  });
+  useBuildHotkey("redoReviewMark", () => apply(history?.redo() ?? null), {
+    enabled,
+  });
```
gregberge: How can the history be optional here? Not sure to understand where it's optional.

apps/frontend/src/pages/Build/BuildReviewState.tsx
```diff
@@ -466,11 +524,124 @@ export function BuildReviewStateProvider(props: {
...
-        {props.children}
+        <BuildReviewHistoryContext value={history}>
+          {props.children}
+        </BuildReviewHistoryContext>
       </BuildReviewAPIContext>
     </BuildReviewStateContext>
   );
 }
+
+function useBuildReviewHistoryValue(args: {
```
gregberge: It's a bad naming, applying diff is not related to history. Globally it's all about the build review state and in this state we can apply / undo / redo.

tests/build.spec.ts
```diff
@@ -353,3 +353,112 @@ loggedTest(
...
+
+    // The digit row is shifted on an AZERTY keyboard, so the "2" the dialog
+    // names arrives as `Digit2` with shift held — which is what pressing the
+    // key with shift produces here too, whatever character it prints.
+    await page.keyboard.press("Shift+Digit2");
+
+    // Going to the first changed snapshot leaves the overview for a diff.
+    await expect(page).not.toHaveURL(/\/overview$/);
+  },
+);
```
gregberge: Should not be a E2E.

tests/build.spec.ts
```diff
@@ -353,3 +353,112 @@ loggedTest(
...
+
+    const dialog = page.getByRole("dialog", { name: "Keyboard Shortcuts" });
+    await expect(dialog.getByText("Undo last review mark")).toBeVisible();
+    await expect(dialog.getByText("⌘", { exact: true })).toHaveCount(0);
+    await expect(dialog.getByText("⇧", { exact: true })).toHaveCount(0);
+    await expect(
+      dialog.locator("kbd").filter({ hasText: "Ctrl" }).first(),
+    ).toBeVisible();
+  },
+);
```
gregberge: Same not a E2E.

gregberge (approved):

apps/frontend/src/util/os.ts
jsfez: Done — both halves use them now. \`keys\` is \`[MOD, "KeyZ"]\`, and \`checkIsModifiedPressed\` (the ⌘→Ctrl conversion) is replaced by a modifier-name → event-flag lookup, so nothing below \`util/os\` knows which platform it runs on.

Best sign it was the right call: knip then flagged \`isMacOS\` as an unused export. Its only consumer outside the module was that conversion, so it is private now.

apps/frontend/src/containers/Build/BuildHotkeys.tsx
jsfez: Because this PR added the first pair of shortcuts that differ only by shift: ⌘Z and ⌘⇧Z. Shift was demanded but never forbidden, so ⌘⇧Z matched ⌘Z too and every redo also ran an undo. Nothing needed it before because no two hotkeys collided that way.

That said you were right that it grew too much — see the other thread, it is down to one rule now.

apps/frontend/src/containers/Build/BuildHotkeys.tsx
jsfez: Cut down a lot. The bulk of it was a special case for digit codes: the digit row is shifted on AZERTY, so pressing the "1" the dialog names sends \`Digit1\` with shift held, and a strict rule killed 1–7 and 0 on your own keyboard.

Declaring the digits as the character they print — \`"1"\`, like \`"?"\` and \`"["\` already were — removes that special case, and the \`hasDigits\` alt one that existed for the same reason. What is left is a single rule: a hotkey named by a character already carries shift in \`event.key\`, so it must not be demanded twice.

apps/frontend/src/pages/Build/BuildPage.tsx
jsfez: Removed.

apps/frontend/src/pages/Build/BuildPreviousReviewDialog.tsx
jsfez: Agreed — \`markDiffs\`, \`undo\`, \`redo\` and \`clearHistory\` are on \`BuildReviewAPI\` now and the separate history context is gone. This call site reads \`api.markDiffs(branchApprovedDiffs, EvaluationStatus.Accepted)\`.

apps/frontend/src/pages/Build/BuildPreviousReviewDialog.tsx
jsfez: It should not — this was leftover from an earlier round. The call is \`api.markDiffs(...)\` under the \`invariant(api)\` that was already there.

apps/frontend/src/pages/Build/BuildReviewHistoryHotkeys.tsx
jsfez: On a reference build: \`BuildReviewStateProvider\` returns a null API for \`BuildType.Reference\`, since there is nothing to review.

Made it explicit rather than optional-chained — the component returns \`null\` when there is no review state, and the inner one takes a non-null \`api\`. No \`?.\`, no \`enabled\` flag.

apps/frontend/src/pages/Build/BuildReviewState.tsx
jsfez: Renamed and moved: it is \`BuildReviewAPI\` that marks, undoes and redoes now, with the stacks as an internal detail of the provider. The component is \`BuildReviewUndoHotkeys\`.

tests/build.spec.ts
jsfez: Moved to \`containers/Build/hotkeys.test.ts\` — a table of (platform, hotkey, event) → matches, covering the AZERTY digit case, ⌘Z vs ⌘⇧Z, and ⌘ vs Ctrl per platform. 15 cases in 135ms instead of one Playwright run.

The declarations and matcher moved to \`containers/Build/hotkeys.ts\` so the test does not pull React in.

tests/build.spec.ts
jsfez: Same file — it asserts \`displayKeys\` directly under a stubbed \`navigator.platform\`, which is the actual unit. Verified both fail when their fix is reverted.

PR #2527, opened 2026-08-22
Move the snapshot's variants to the top bar

gregberge:
## What

Browser, viewport, color scheme, media type and story mode move out of the right sidebar's Metadata panel and into the build page's top bar, between the snapshot's name and the pane's controls.

The panel keeps everything that describes the test rather than the variant: SDK, automation library, threshold, baseline, retry, story kind, tags, URL, test, annotations and trace.

## Why

Those five are the dimensions a snapshot varies along — what the reviewer is looking at, and the siblings they can jump to — so they belong next to the name, not buried in a panel among facts of a different kind. They also disappeared entirely whenever the sidebar was closed.

**This fixes a bug:** the switch hotkeys (<kbd>B</kbd>, <kbd>V</kbd>, <kbd>M</kbd>) are registered by the chips themselves, so closing the sidebar unregistered them. The toolbar is always mounted, so they now work either way.

## Notes for the reviewer

- **The first commit is a pure move** — the five components leave `sidebar/metadata/` (they are no longer sidebar rows) for `metadata/variants/`, renamed after what they do, and `utils.ts` moves up to `metadata/`, which both sides import from. The only behavioural change is dropping the `MetadataRow` wrapper, which was sidebar row styling. Display rules are untouched: single values render as static chips, color scheme appears only when a dark sibling exists, media type only for print.
- **The second commit fixes ordering**, which the move made more visible. Each switcher collects its values by walking the sibling diffs, so without a sort the chips came out in whatever order the diff list handed over. Browser and color scheme had no sort at all; viewport sorted on width alone, leaving same-width pairs (the seeds have 375×720 and 375×1440) in arrival order. They now follow the orders the filter menu and the build overview already use for the same values — note that means Dark before Light, as the overview already shows.
- `DetailToolbarTitle` gains a `min-w-40` floor. With chips on the row the title was the only flexible item and collapsed to `bear-2560×` at a 1024px window; past the floor the row wraps instead. This primitive is shared with the media viewer, whose own code already anticipated wrapping.
- **Argos will show reordered/moved chips across the `build-*` baselines** — that is the point of the change, but it means a broad approval.

## Verification

- `pnpm run static-checks` — all 11 tasks pass, knip included.
- `tests/build.spec.ts` (21) and `tests/media.spec.ts` (13) pass on chromium.
- New `metadata/utils.test.ts` (6 cases) feeds each helper its values in the wrong order on purpose and asserts the output does not depend on it. It caught a real mistake while being written: the numeric version compare was right where the expectation was wrong.
- Checked in the running app: static-chip diffs, a markdown snapshot (no chips at all), a removed diff (metadata read from the baseline), and — by temporarily renaming screenshots in the test DB, since no seed produces them — the multi-value switchers, confirming `aria-current`, the navigation links, and that a group arriving WebKit/Chromium/Firefox renders Chromium, Firefox, WebKit.


jsfez:
I love the idea of this PR. 
Are you sure the title is not cropped on an average screen size? 
Can you share a screenshot of a screen with several selectors displayed in the header? Or maybe add the seeds to see it in Argos.

jsfez (approved):

gregberge:
Good catch — **yes, it was cropped**, at 1280 and 1366. Thanks for pushing on it; the screenshots you asked for are in the Argos comment above (before/after pair), and there's now a seeded build so this stays visible.

## The title was cropped, and `min-w-40` was the wrong fix

I measured a 76-character name against a full set of chips, with both sidebars open:

| Viewport | Title box | Clamped |
| :-- | :-- | :-- |
| 1280 | 196px | 20px (3 lines' worth in a 2-line box) |
| 1366 | 228px | 20px |
| 1440 | 302px | — |

`line-clamp-2` swallows the third line with no ellipsis, so it read as a complete name that just happened to end at "and".

The cause is more interesting than a missing minimum: **a flex line is broken up on base sizes**, and `flex-1` means `flex-basis: 0`. The title counted for nothing when the browser decided what fits on the row, so the row never wrapped on its account — it just handed the name whatever the chips and controls left over. A `min-width` can't fix that, because a minimum can't make a line wrap; it can only push the other items or overflow the row. I'd added `min-w-40` for exactly this and it was papering over the mechanism.

Giving the title a real basis (`basis-88`) lets it break the line: the chips drop to their own row when the name needs the space, and the name still grows into whatever they leave. Measured across 1024 → 1920, nothing is clamped and neither the row nor the page overflows.

## Seeded, so it's visible in Argos

You were right that seeds are the better answer than a screenshot. No seeded build had sibling diffs to switch *between*, so the multi-value switchers couldn't be seen at all — I'd only been able to check them by hand-editing rows in a local database, which is exactly the kind of thing that rots. `createVariantSwitchersScenario` seeds the shape that produces them: one snapshot across two browsers and two viewports, which `getVariantKey` folds into one key via the browser prefix and the ` vw-<width>` suffix. The name is long on purpose.

The spec asserts the row seats it without cropping, since that's the part nobody can eyeball. I checked it fails on the layout this branch replaces before trusting it.

## Two other things that came out of this

- The chip group had no accessible name — in the sidebar it sat under a "Metadata" heading, out here it was a bare row of links. Now `role="group"` + a label, which is also what let the spec scope its locators (the title beside it is a link whose name starts with "chromium/", so an unscoped `Chromium` matched both).
- Expect **media page baselines to move too**: `DetailToolbarTitle` is shared with the media viewer, so its toolbar wraps at narrow widths now rather than crushing the file name. Its own code already anticipated that ("even when the row is too narrow to hold them and they wrap under the title").

Full suites pass (`build.spec.ts` 21 + the new one, `media.spec.ts` 13), and `static-checks` is green.

gregberge:
CI is settled — everything functional passes (e2e on all 3 shards, integration, unit, storybook, static analysis). `argos/argos` reads as failing only because it is [`changes-detected`](https://app.argos-ci.com/argos-ci/argos/builds/5708) awaiting review.

Correcting my own overstatement in the PR body — I said to expect "essentially all `build-*` baselines" to move. The real footprint is **20 changed + 2 added out of 247**, which is 11 distinct snapshots across the two browser projects:

- **`build-variant-switchers`** — added, the new seeded scenario.
- **7 `build-*` snapshots** — the ones showing a snapshot with the toolbar open (`changes detected`, `stable`, `stable with fail screenshots`, `rejected`, `reference`, `orphan`, `subset`). Every `build-overview-*` is untouched, since the overview has no toolbar.
- **`media-share-page` and `media-share-sidebar`** — the shared-title change I flagged, showing up exactly where predicted.

So the review is 11 snapshots, not a blanket approval. The media pair is the one worth a second look, since that page was not the target of this PR.

gregberge:
Both of these are in, with before/after pairs in the Argos comment above.

**Build toolbar — the wrapped row now holds to the right.** `justify-between` has nothing to space a lone item against, so once the chips took the first line the controls dropped to the left edge under the name, which is the gap your arrow was pointing at. `ml-auto` fixes it, and it's the same thing `MediaViewToolbar` already does for its own controls when its row wraps. Screenshots are at 1180px, where the row wraps.

**Media page — the numbers moved to the sidebar.** The dimensions and file size now follow the description, as one block: the numbers say what the prose is about, and the gap collapses on its own when there is no description, so they're simply the first thing in the column instead (checked on the video, which has no description).

Two notes on what that dragged in:

- I moved the **state** label (`before` / `after`) along with them, since `getMediaFacts` produced all three as one line and leaving one word behind would have kept the whole slot alive. It still only appears when the second pane isn't up — and the condition got slightly more honest in the move: it's now "a pair is actually on screen", which a video never is whatever it was paired with. Say the word if you'd rather it stayed in the bar.
- That was `DetailToolbarTitle`'s only `meta` caller, so **the slot is gone** and the title stops being two things sharing a baseline. It's also the last of the crowding that started this thread: nothing but the name is in that half of the row now.

`build.spec.ts` + `media.spec.ts` pass (35), `static-checks` green. `media-share-page` and `media-share-sidebar` baselines move again, this time on purpose.

gregberge:
Both were the same root cause, and you were right on both counts. Before/after pairs are in the Argos comment above.

**The media page had no business wrapping.** I had the name asking for a flat `22rem` of the row during line measurement, which is far more than `checkout.png` needs — it reserved 352px, pushed the controls to a second line, and left two thirds of the row empty. The name now asks for its own length, so a short one leaves the row alone.

**Left at that, a long name is the opposite problem:** it asks for all 563px it would like, takes the line to itself, and folds the chips *and* the controls underneath together — which is worse than what you were looking at. So the ask is capped only where the row carries something besides the name and its controls (`crowded`, which the build page sets and the media page does not).

**Chips now fold under the controls, not the reverse.** A flex row folds from the end, so the slot that goes under is whichever is last — the chips now. That's the right way round for the reason you'd expect: the controls are identical on every snapshot and should keep their place, while the chips are as many as the snapshot happens to have. One cost worth naming: **on a single line the chips now read after the controls rather than before them**, which reverses the placement you picked earlier. There's no way to have both — the item that folds is the item that comes last. Say so if you'd rather keep chips-then-controls on one line and accept the controls folding.

**Gap tightened** to `gap-y-2`. A full `gap-4` between rows read as two bars rather than one folded over.

Measured on the seeded long name, both sidebars open:

| Width | Rows | Layout |
| :-- | :-- | :-- |
| 1280 | 2 | name + controls, then chips under |
| 1440 | 1 | everything on one line |
| 1920 | 1 | everything on one line |

Nothing is clamped at any width, and the media page is one line at 1280 and 1440. The old `22rem` wrapped at 1440 too, which is the unnecessary wrap you spotted.

`build.spec.ts` + `media.spec.ts` pass (35), `static-checks` green.

gregberge:
Redesigned per the latest feedback — screenshots in the Argos comment above (1440 one-line, 1100 folded).

**Switchers only where switching is possible.** A dimension with a single value renders nothing in the toolbar; the facts are back in the Metadata panel as plain chips (browser + version, viewport, Dark, Print, story mode) — reading the current snapshot only, no sibling walking, so they're much simpler than the rows that left. Media type is gone from the toolbar entirely: a variant key never differs by it, so it can never be switched.

**Buttons, not chips** — same height and surface as the pane controls beside them (`LinkButton variant="secondary"`, medium). Compact per your suggestion: browsers are their logos alone, viewports their width alone, color schemes a sun against a moon; tooltips and accessible names carry the full story. The active segment says `aria-current="page"`, and the global `on:` variant now answers it alongside pressed/selected/expanded — note this also gives **Pagination's active page** the on-look it arguably always deserved (its baseline may shift slightly).

**Variants left of the controls, folding under them.** `flex-wrap-reverse` on the right-side cluster: items fill lines in order (switchers first), reversed stacking puts that first line at the bottom — so one line reads name · switchers · controls, and when the row runs out the controls keep the top line and the switchers drop below, right-aligned. The cluster grows ahead of the title but never past its content, so free space unfolds the switchers before it lengthens the name. Measured on the seeded build: one line at 1280/1440/1920 (the compact buttons are 177px where the chips were 342), fold below ~1150, title never clamped, no overflow at any width down to 1024.

`build.spec.ts` + `media.spec.ts` pass (35) — the switcher spec now asserts "Dark" is *not* in the toolbar and *is* in the panel — `static-checks` green.

PR #2529, opened 2026-08-23
feat(staff): add a Revenue page, read from a Stripe invoice mirror

jsfez:
A staff-only Revenue page: what Argos invoiced, month by month. Read from the Stripe invoices themselves rather than recomputed from usage, so the amounts carry what Stripe actually charged — negotiated prices, coupons, credit notes — instead of a second pricing engine of ours that would have to agree with the first. Figures are ex-tax, net of credit notes, and stated in euros, dollar invoices converted at a fixed rate.

## What the page shows

Three cards — last month, the running month, and the change between them — each breaking down into its monthly, yearly and GitHub Marketplace parts. Under them a stacked chart of the year, then two tables:

- **Monthly plans**, month by month, with the change, the teams invoiced and the ARPU. Every row expands into the teams behind it, each with what it was invoiced in its own currency and in euros, and a link to its Stripe customer.
- **Annual contracts**, one row per contract, with the invoices its figure is read from — including the contracts that contributed nothing, which a total alone would hide. Invoices raised but unpaid are counted and flagged, so a contract in collection stays visible.

## Where the figures come from

A `stripe_invoices` table mirrors the fields the figures need, so a view costs a few queries instead of a paginated walk of a third party.

Webhooks are treated as signals only: every invoice or credit-note event re-reads the invoice from the API. Stripe guarantees no delivery order, retries for days, and embeds at most ten lines per payload — trusting the payload would let a stale event regress a paid invoice to open, or truncate the period an annual bill is recognized by. A daily cron reconciles what the webhooks may have missed — new invoices, invoices whose credit notes landed since, rows still unsettled from before the window — and records the stretch it covered. The page refuses a window the mirror was never backfilled for, rather than reporting zeros that would read as figures.

Three things the arithmetic does that a simpler reading gets wrong:

- Credit notes are netted **ex-tax**, read from the notes themselves, so a refunded taxed invoice nets to zero instead of minus its VAT.
- Annual contracts are **amortized day by day** over the months they cover, so a renewal weighs on the months it pays for, an upsell on the stretch it was sold for, and a month's figure never moves once written.
- An annual bill **never lands in a month's monthly figures**, whoever it belongs to now, so a churned yearly team cannot spike the month its renewal was raised in.

GitHub Marketplace is reported beside the Stripe figures, never inside them: GitHub exposes no seller invoice API, so it is priced from the listing's plans times the active marketplace subscriptions.

## Known limits

- **The Marketplace figure is today's book, drawn flat across history.** Every past month is charted at the current subscriptions times the current list price, so updating either rewrites the historical band. The tooltip says as much, but pricing each month from the subscriptions active _that_ month would end it.
- **The "Month over month" card compares a partial month to a complete one**, so it reads as a fall for most of the month. Deliberate — the running figure is what staff want to watch as it fills, and the card's hint says as much.
- **`EUR_PER_USD` is a hand-maintained constant** (0.855, ECB on 2026-08-21). Deliberate — the figures should move with the book, not the market — but it drifts and nothing reminds anyone to update it.

## Deploying

Order matters. **Enable the webhook events only after the deploy is live**: the currently-deployed handler throws on any event type it does not know, and the route turns that into a 500, so events enabled early would be retried for days and could get the endpoint disabled — taking the existing subscription webhooks down with it.

1. Deploy (the migration adds `stripe_invoices`, `stripe_invoice_syncs`, and a price column on `plans`).
2. On the Stripe webhook endpoint, enable exactly: `invoice.created`, `invoice.updated`, `invoice.finalized`, `invoice.paid`, `invoice.voided`, `invoice.marked_uncollectible`, `invoice.deleted`, `credit_note.created`, `credit_note.updated`, `credit_note.voided`.
3. Backfill the mirror once: `stripe/bin/sync-stripe-invoices 760`. The recorded sweep is what lets the page report that far back; the run is sequential and idempotent, so a failed one can simply be run again, and a partial one leaves the page refusing the window rather than showing a truncated figure.
4. Price the marketplace plans: `github/bin/sync-github-marketplace-prices`, or wait for the daily cron — that figure reads €0 until it has run.

The page stays empty-handed but honest until step 3 finishes: it reports itself unavailable with the reason, which is also what it does in the e2e suite.

### DIFF
--- a/apps/backend/db/migrations/20260823065836_stripe-invoices.js
+++ b/apps/backend/db/migrations/20260823065836_stripe-invoices.js
@@ -4,7 +4,13 @@
  * Kept in sync by the Stripe webhooks plus a reconciliation sweep, so the page
  * costs a query instead of a paginated walk of Stripe on every view. Only the
  * fields the revenue arithmetic reads are mirrored; the invoice itself stays
- * in Stripe. Amounts are in the currency's minor unit, as Stripe states them.
+ * in Stripe. Amounts are in the currency's minor unit, as Stripe states them —
+ * 32-bit on purpose: the cap is ~21M in minor units, three orders of magnitude
+ * above the largest invoice ever raised, and staying `integer` keeps the
+ * values plain numbers where `bigint` would come back from pg as strings.
+ *
+ * `stripe_invoice_syncs` records each completed sweep, so the reader can tell
+ * a window the mirror never covered from a quiet one.
  *
  * @param {import('knex').Knex} knex
  */
@@ -25,12 +31,22 @@ export const up = async (knex) => {
     table.integer("total").notNullable();
     table.integer("totalExcludingTax");
     table.integer("totalTaxesAmount");
-    table.integer("prePaymentCreditNotesAmount").notNullable();
-    table.integer("postPaymentCreditNotesAmount").notNullable();
+    // The credit notes' ex-tax totals added up — ex-tax like the base they are
+    // taken off, where the invoice's own rollup fields are tax-inclusive.
+    table.integer("creditedAmountExcludingTax").notNullable();
     // The longest stretch one of the invoice's lines covers, resolved at
     // ingest — what tells an annual bill from a true-up.
     table.dateTime("periodStart");
     table.dateTime("periodEnd");
+  });
+
+  await knex.schema.createTable("stripe_invoice_syncs", (table) => {
+    table.bigIncrements("id").primary();
+    table.dateTime("createdAt").notNullable();
+    table.dateTime("updatedAt").notNullable();
+    // The window the sweep read, and when it finished reading it.
+    table.dateTime("sinceDate").notNullable();
+    table.dateTime("completedAt").notNullable();
   });
 };
 
@@ -38,5 +54,6 @@ export const up = async (knex) => {
  * @param {import('knex').Knex} knex
  */
 export const down = async (knex) => {
+  await knex.schema.dropTable("stripe_invoice_syncs");
   await knex.schema.dropTable("stripe_invoices");
 };
--- a/apps/backend/db/migrations/20260823065836_stripe-invoices.js
+++ b/apps/backend/db/migrations/20260823065836_stripe-invoices.js
@@ -1,0 +1,42 @@
+/**
+ * Mirror of the Stripe invoices the revenue page reads.
+ *
+ * Kept in sync by the Stripe webhooks plus a reconciliation sweep, so the page
+ * costs a query instead of a paginated walk of Stripe on every view. Only the
+ * fields the revenue arithmetic reads are mirrored; the invoice itself stays
+ * in Stripe. Amounts are in the currency's minor unit, as Stripe states them.
+ *
+ * @param {import('knex').Knex} knex
+ */
+export const up = async (knex) => {
+  await knex.schema.createTable("stripe_invoices", (table) => {
+    table.bigIncrements("id").primary();
+    table.dateTime("createdAt").notNullable();
+    table.dateTime("updatedAt").notNullable();
+    table.string("stripeInvoiceId").notNullable();
+    table.unique("stripeInvoiceId");
+    table.string("stripeCustomerId").notNullable().index();
+    table.string("stripeSubscriptionId");
+    // Stripe's own `created`, which is what files an invoice into a month.
+    table.dateTime("stripeCreatedAt").notNullable().index();
+    table.string("status").notNullable();
+    table.string("billingReason");
+    table.string("currency").notNullable();
+    table.integer("total").notNullable();
+    table.integer("totalExcludingTax");
+    table.integer("totalTaxesAmount");
+    table.integer("prePaymentCreditNotesAmount").notNullable();
+    table.integer("postPaymentCreditNotesAmount").notNullable();
+    // The longest stretch one of the invoice's lines covers, resolved at
+    // ingest — what tells an annual bill from a true-up.
+    table.dateTime("periodStart");
+    table.dateTime("periodEnd");
+  });
+};
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+export const down = async (knex) => {
+  await knex.schema.dropTable("stripe_invoices");
+};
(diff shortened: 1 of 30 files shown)


apps/backend/db/migrations/20260823065836_stripe-invoices.js
```diff
@@ -0,0 +1,42 @@
+/**
+ * Mirror of the Stripe invoices the revenue page reads.
```
gregberge, line 2: If you do this, I think it would be nice to show it to users too in their dashboard, the last invoice and the next invoice to come?

jsfez: yes

apps/backend/src/github/bin/sync-github-marketplace-prices.ts
```diff
@@ -0,0 +1,22 @@
...
+const main = callbackify(async () => {
+  const count = await syncGithubMarketplacePlanPrices();
+  logger.info(`${count} plans priced from the Marketplace listing`);
+});
+
+main((err) => {
+  if (err) {
+    throw err;
+  }
+});
```
gregberge: I think `await` at top level should be available now?

apps/backend/src/stripe/bin/sync-stripe-invoices.ts
```diff
@@ -0,0 +1,33 @@
...
+  const since = new Date(Date.now() - days * 24 * 3600 * 1000);
+  const count = await syncStripeInvoices({ since });
+  logger.info(`${count} invoices mirrored since ${since.toISOString()}`);
+});
+
+main((err) => {
+  if (err) {
+    throw err;
+  }
+});
```
gregberge: Same remark, top level await?

apps/backend/src/processes/proc/worker.ts
```diff
@@ -39,6 +42,26 @@ scheduleCron("media-retention", "15 * * * *", (context) =>
   purgeExpiredMedia(context.date),
 );
 
+// The safety net under the invoice webhooks: re-reads a window wide enough to
+// catch anything a missed delivery left behind. Daily, because the webhooks
+// are the live path — the sweep only has to beat an operator noticing.
+scheduleCron("stripe-invoice-sync", "45 4 * * *", async (context) => {
```
gregberge, line 48: Why do we use a cron and not webhooks for this?

gregberge (approved):

apps/backend/src/processes/proc/worker.ts
jsfez: We use webhooks, with a cron job as a fallback to catch any events we might miss—for example, if our server is down. However, Stripe may already retry webhook deliveries in that case.

apps/backend/db/migrations/20260823065836_stripe-invoices.js
jsfez: Worth splitting in two, because the mirror helps with one half and not the other.

**Last invoice** is nearly free: the row already holds the date, the amount ex-tax, the currency, the status and what credit notes gave back. What it deliberately does not hold is anything only a customer needs — the hosted invoice URL, the PDF, the line items, the human-facing invoice number — so a dashboard view would mean widening the table (cheap) rather than a new source.

**Next invoice** is not in the mirror at all, by design: Stripe's upcoming invoice is a preview, it carries no id, and `invoice.upcoming` is one of the event types ingestion drops on purpose. Showing it would be a live `invoices.retrieveUpcoming` per account, or mirroring the subscription's `next_payment_attempt` — a different mechanism either way.

So: last invoice is a follow-up on this table, next invoice is its own piece of work.

apps/backend/src/github/bin/sync-github-marketplace-prices.ts
jsfez: You're right, and there is already precedent for it: `src/database/bin/run-seed.ts` is plain top-level await with no wrapper. Both scripts here copied the `callbackify` shape from `src/build/bin/queue-pending-builds.ts`, which predates it.

Nothing is lost by dropping it either — the wrapper exists to make the process exit non-zero on failure, and an unhandled rejection from a top-level await does that on its own since Node 15. I'll simplify both.

apps/backend/src/stripe/bin/sync-stripe-invoices.ts
jsfez: Same here — simplifying both to top-level await, like `src/database/bin/run-seed.ts` already does.

apps/backend/src/processes/proc/worker.ts
jsfez: To complete Jeremy's answer: webhooks are the live path — every `invoice.*` and `credit_note.*` event re-reads the invoice from the API. The cron is there for what webhooks structurally cannot cover, which is more than downtime:

- Stripe retries a failing endpoint for about three days and then stops, so any outage longer than that loses those events for good.
- An event type that is not enabled on the endpoint is never delivered at all — including one added later, whose backlog no retry will replay.
- Some states change without an invoice event reaching us: a credit note issued on an old invoice, or an invoice left `open` that is paid or voided long after. The sweep re-reads those two sets on purpose, not just the window's new invoices.
- The sweep records the window it covered, and the page refuses to report a window no sweep has covered. Without the cron the reader would go dark rather than answer from a mirror nothing reconciles.

So the cron is not a duplicate of the webhooks; it is what makes the mirror provable rather than merely current.

apps/backend/src/github/bin/sync-github-marketplace-prices.ts
jsfez: Done in 370716c4 — both scripts now await at the top level. The explicit `process.exit(0)` stays: each holds a Redis connection that keeps the event loop alive, which is why `run-seed.ts` and `generate-openapi.ts` end the same way.

apps/backend/src/stripe/bin/sync-stripe-invoices.ts
jsfez: Done in 370716c4 — both scripts now await at the top level. The explicit `process.exit(0)` stays: each holds a Redis connection that keeps the event loop alive, which is why `run-seed.ts` and `generate-openapi.ts` end the same way.

PR #2534, opened 2026-08-23
fix(auth): stop writing "undefined" as the OAuth client id

jsfez:
The `github.loginUrl` and `gitlab.loginUrl` config defaults interpolated `process.env` at module load, behind convict's own env plumbing:

```js
default: `https://github.com/login/oauth/authorize?scope=user:email&client_id=${process.env["GITHUB_CLIENT_ID"]}`,
```

Whenever the variable is absent, the template renders the string `"undefined"` and the login button points at `…/authorize?client_id=undefined`, i.e. GitHub's 404 page. That is the case for a self-hosted install without a GitHub app, for the E2E web server in CI, and for `pnpm run dev:prod-ro`, which deliberately loads no `.env` — where this was found.

The two URLs are now built where they are served, from the client id convict resolved.

## Checks

Built server, `/config.js` and `/auth/google/login`:

| | before | after (no credentials) | after (credentials set) |
|---|---|---|---|
| GitHub | `client_id=undefined` | `client_id=` | `client_id=Iv1.testclientid` |
| GitLab | `client_id=undefined` | `client_id=` | `client_id=gitlab-test-app-id` |
| Google | `client_id=` | unchanged | `client_id=1234-test.apps.googleusercontent.com` |

Google never had the bug — its client id already went through convict properly — and is listed only to show the three paths now agree.

`check-types`, `lint` and `check-format` pass.

## Out of scope

An instance with no OAuth app still renders "Continue with GitHub"; the button now leads to a missing-`client_id` error instead of a 404. Hiding a provider the instance cannot authenticate with is a larger change (client config, login and signup screens, account authentication cards, pinned client ids for the E2E screenshots) and is deliberately left out.

### DIFF
--- a/apps/backend/src/config/index.ts
+++ b/apps/backend/src/config/index.ts
@@ -297,10 +297,6 @@ export function createConfig() {
         default: "https://github.com/apps/argos-ci-dev",
         env: "GITHUB_APP_URL",
       },
-      loginUrl: {
-        format: String,
-        default: `https://github.com/login/oauth/authorize?scope=user:email&client_id=${process.env["GITHUB_CLIENT_ID"]}`,
-      },
       webhookSecret: {
         format: String,
         default: "development",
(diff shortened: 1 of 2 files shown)


gregberge (approved):
I think it makes more sense like this yes.

PR #2537, opened 2026-08-24
chore(backend): let prod-ro sign in with GitHub, without writing

jsfez:
## Why

`pnpm run dev:prod-ro` offers "Continue with GitHub", and the button cannot work: the mode carries no git-provider credentials, so the authorize URL has no client id and GitHub answers 404. An email code read out of local Redis was the only way in.

## What it does

Under `ARGOS_TARGET=prod-ro`, `/auth/github` resolves the account the profile already belongs to and opens the session:

```ts
if (config.get("target") === "prod-ro") {
  const { data: profile } = await octokit.users.getAuthenticated();
  return getUserAccountFromGithubId({ githubId: profile.id });
}
```

Everything the normal path does beyond that is a write the `argos_dev_ro` role does not have — refreshing the stored OAuth token, creating the account when it is missing, joining SSO teams. The first is the one CONTRIBUTING warned about: it would replace a production access token with one minted by the local OAuth app, and since production spends that token on `apps.listInstallationsForAuthenticatedUser`, the user's "connect a repository" screen on `app.argos-ci.com` would go empty until they signed in there again. `user_sessions` and `team_users.lastAuthMethod`, the two tables the mode grants, are the only rows touched.

### Why a dedicated resolver

An earlier revision reused `getOrCreateUserAccountFromGhAccount` with a read-only fetch, which is write-free for a complete, already-linked account — and only for that. Four ordinary row shapes still reach a write, each surfacing as a raw `permission denied` 500 rather than a usable message:

| row shape | what the shared resolver does |
| --- | --- |
| account matched by email, not yet linked | patches `accounts.githubAccountId` |
| user row with no email | inserts `user_emails`, accepts pending invites |
| soft-deleted user | patches `deletedAt` |
| no user at all (an org carried by a team) | creates an account |

It also invariants on `model.emails`, which is nullable on `github_accounts` — every row created by a webhook rather than a login has it null, so an org member signing in would get a 500 before reaching any error path. `getUserAccountFromGithubId` is one indexed query and one 403, and it makes "this login writes nothing" readable in one place.

## Where the credentials come from

From your `.env`, with nothing to set up. `selectDotEnv` decides what a target may take from that file: everything for a normal local process, exactly `GITHUB_CLIENT_ID` and `GITHUB_CLIENT_SECRET` for prod-ro. The rest — Resend, Stripe, the GitHub app private key — stays absent, which is what keeps browsing production data side-effect free.

Referencing the pair from `.env.prod-ro` instead was tried and dropped: it needs two fields in a 1Password vault most of us cannot write to, so until someone with access adds them the mode does not start at all — a regression for everyone, for credentials every developer already has on disk.

Exporting the variables still overrides the file, hence the two names in `turbo.json`: turbo runs in `envMode: strict`, so an override that is not declared there is filtered out between `op run` and the server.

## About `GITHUB_CLIENT_SECRET` leaving the guardrail list

This is the one change that narrows a guardrail, and the obvious justification for it is wrong, so it is worth stating precisely. "The token is scoped to `user:email`, so it cannot write" holds for an OAuth App — Argos' integration is a **GitHub App**, where the requested scope does not bound a user-to-server token: it carries the app's installation permissions. The credential could therefore write *through the app*.

What makes it safe here is the use, not the credential: prod-ro spends the token on a single profile read and never stores it. That obligation is written next to the list, so the entry goes back the day it stops holding. `GITHUB_APP_PRIVATE_KEY`, which needs no user at all to post comments and statuses, stays listed, as do the GitLab and Google secrets.

## Testing

- `config/dotenv.test.ts` pins the allow-list: everything passes through for a local target, prod-ro gets the two names and nothing else, and each excluded secret is checked by name.
- `services/prod-ro-github-login.e2e.test.ts` covers the resolver: it returns the user account behind a profile, refuses an unattached profile, and refuses a GitHub organization carried by a team account — that last one fails without the `userId` guard, verified by removing it.
- The flow was exercised end to end against production from a local prod-ro process: the button reaches GitHub's consent screen and the session opens.
- `pnpm run static-checks` passes.

The callback branch itself has no automated test — reaching it needs a server booted in prod-ro against production, which no test environment has.

## Known, not addressed here

The login screen still renders the GitLab and Google buttons in prod-ro, where they have no client id and lead to a provider error page. Hiding a provider the instance cannot authenticate with is a frontend change of its own (login screen, signup screen, account cards) and is deliberately left out.

### DIFF
--- a/.env.prod-ro
+++ b/.env.prod-ro
@@ -8,6 +8,9 @@
 # (Resend, Stripe, git-provider apps… their absence is what keeps browsing
 # production data side-effect free). See CONTRIBUTING.md § Running against
 # production data.
+#
+# A gitignored .env.prod-ro.local is loaded after this one and wins, for the
+# op:// references that differ from vault to vault.
 ARGOS_TARGET=prod-ro
 
 # No password in the URL: PG_IAM_AUTH mints a short-lived RDS IAM token per
--- a/.env.prod-ro
+++ b/.env.prod-ro
@@ -18,6 +18,13 @@ PG_IAM_AUTH=true
 
 SQIDS_ALPHABET=op://argos-dev/argos-prod-ro/SQIDS_ALPHABET
 
+# The dev GitHub OAuth app, for logging in. Only the login credentials are
+# here: the *app* private key, which posts comments and statuses, stays out —
+# and the callback writes nothing in this mode, so which app signs the user in
+# no longer decides what lands in the production rows.
+GITHUB_CLIENT_ID=op://argos-dev/argos-prod-ro/GITHUB_CLIENT_ID
+GITHUB_CLIENT_SECRET=op://argos-dev/argos-prod-ro/GITHUB_CLIENT_SECRET
+
 # Production screenshots (public CDN, read via s3:GetObject only).
 AWS_SCREENSHOTS_BUCKET=argos-ci-production
 S3_PUBLIC_IMAGE_BASE_URL=https://files.argos-ci.com/
--- a/apps/backend/src/database/services/prod-ro-github-login.e2e.test.ts
+++ b/apps/backend/src/database/services/prod-ro-github-login.e2e.test.ts
@@ -1,0 +1,48 @@
+import { invariant } from "@argos/util/invariant";
+import { beforeEach, expect, test } from "vitest";
+
+import { GithubAccount } from "@/database/models";
+import { factory, setupDatabase } from "@/database/testing";
+
+import { getUserAccountFromGithubId } from "./account";
+
+beforeEach(async () => {
+  await setupDatabase();
+});
+
+async function getGithubId(account: { githubAccountId: string | null }) {
+  invariant(account.githubAccountId, "account has no GitHub account");
+  const ghAccount = await GithubAccount.query().findById(
+    account.githubAccountId,
+  );
+  invariant(ghAccount, "GitHub account not found");
+  return ghAccount.githubId;
+}
+
+test("resolves the user account behind a GitHub profile", async () => {
+  const userAccount = await factory.UserAccount.create();
+  const githubId = await getGithubId(userAccount);
+
+  const result = await getUserAccountFromGithubId({ githubId });
+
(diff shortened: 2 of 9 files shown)


apps/backend/src/database/services/prod-ro-github-login.e2e.test.ts
gregberge, line 1: Remove this test.

apps/backend/src/web/api/auth.ts
```diff
@@ -110,6 +112,24 @@ router.use(
       token: result.access_token,
       proxy: false,
     });
+
+    // Read-only login. Everything the normal path does beyond opening the
+    // session is a write the prod-ro role does not have, so the profile can
+    // only be matched against an account that already exists.
+    if (config.get("target") === "prod-ro") {
```
gregberge, line 119: We will not do custom code for this.

.env.prod-ro
```diff
@@ -8,6 +8,9 @@
 # (Resend, Stripe, git-provider apps… their absence is what keeps browsing
 # production data side-effect free). See CONTRIBUTING.md § Running against
 # production data.
+#
+# A gitignored .env.prod-ro.local is loaded after this one and wins, for the
+# op:// references that differ from vault to vault.
```
gregberge, line 13: Remove it I don't even understand what does it mean.

package.json
```diff
@@ -4,7 +4,7 @@
   "type": "module",
   "scripts": {
     "dev": "turbo run watch-build watch-server watch-codegen --concurrency 100",
-    "dev:prod-ro": "op run --env-file=.env.prod-ro -- pnpm run dev",
+    "dev:prod-ro": "op run --env-file=.env.prod-ro $([ -f .env.prod-ro.local ] && echo --env-file=.env.prod-ro.local) -- pnpm run dev",
```
gregberge, line 7: Why do we want to support local? Let's keep the same file for everyone.

gregberge (changes requested):

PR #2538, opened 2026-08-24
feat(billing): show a team its invoices, and the one to come

jsfez:
Greg asked, on #2529, that the invoices be shown to customers too. Two halves, and the mirror only helps with one.

## The invoices already billed

They come straight from the mirror, widened with the three fields it deliberately left out — `number`, `hostedInvoiceUrl`, `invoicePdfUrl`. Without them the list is one a customer cannot download from, which sends them to the Stripe portal anyway; with them the page stays a single indexed query, and the `invoice.*` webhooks already refresh every row.

## The invoice to come

Not in the mirror, by design: Stripe's upcoming invoice is a preview, it carries no id, and `invoice.upcoming` is one of the event types ingestion drops. So it is read live, per view, from `invoices.createPreview` — no Redis cache.

Computing it here from `flatPrice` plus the metered screenshots would have been cheaper and wrong. Previewing a real subscription in test mode returns:

```
1 × Pro plan (at $100.00 / month)                     $100.00
1 × Add-on: GitHub SSO (at $50.00 / month)             $50.00
0 × Additional screenshot (at $0.004 / month)           $0.00
0 × Additional Storybook screenshot (at $0.0015 / mo)   $0.00
                                              total   $150.00
```

The local estimate would have announced $100: the add-on exists nowhere in our database, and coupons, proration and tax move the figure the same way.

## The page

One list, the invoice to come as its first row — laid out after Vercel's invoices page. That row has its own Suspense boundary, so the history renders at once and only the first row waits on Stripe. A Stripe that answers badly takes down that row and nothing else; when there is genuinely nothing to come — a trial with no card, a subscription ending, a team GitHub bills — the row does not render at all.

Settings › Billing keeps the plan, the consumption, the add-ons and spend management, and gains a card linking to the page. Admins only, enforced in the resolver as well as on the route.

## Deploying

Run `sync-stripe-invoices 1200` once. The nightly sweep only re-reads a shallow window plus the unsettled invoices, so older mirrored rows keep a null number and no PDF link until a deep window has been swept.

## Test plan

- `tests/account-invoices.spec.ts` — the history (order, numbers, statuses, amounts, the two links behind the row menu, new Argos baseline `account-invoices`), the way in from Settings › Billing, and a non-admin member getting a 404.
- The Stripe preview was exercised against test mode, both against a live subscription and against two trials with no card on file — those return `404 invoice_upcoming_none`, which the service reads as "no next invoice" rather than as a failure.
- `pnpm run static-checks` passes. The failures left in `test:unit` / `test:integration` are the S3 suites failing on expired AWS credentials, unrelated to this branch.


jsfez:
Découpée en deux, pour que le miroir soit déployé et rempli avant que la page n'existe — sans quoi un client verrait une liste dont seules les factures récentes sont téléchargeables :

- #2569 — l'élargissement du miroir seul, à déployer puis `sync-stripe-invoices 1200`
- #2570 — la page, basée sur #2569

Et #2571 pour le correctif de flake sur les baselines passkeys, qui n'avait de rapport avec ni l'une ni l'autre.

PR #2543, opened 2026-08-24
fix(build): update the text diff when another snapshot is selected

jsfez:
Reported by Sebastien Lorber on [build 1903](https://app.argos-ci.com/meta-open-source/docusaurus/builds/1903/422242813) of `meta-open-source/docusaurus`: clicking another text snapshot changed the URL and the sidebar selection, but the diff in the middle kept showing the previous snapshot. A full browser reload was needed to see the right one.

## Cause

`@pierre/diffs` caches a rendered diff — hunks *and* highlight — under its `cacheKey`, and falls back to the **file name** when none is given. Every file we hand it is named `"snapshot"`, so every text diff in the app shared the key `"snapshot:snapshot"`.

`DiffHunksRenderer.renderDiff` then computes `newContent = !areDiffTargetsEqual(diff, renderCache.diff)`, and `areDiffTargetsEqual` treats two diffs as equal as soon as their cache keys match. The incoming diff was reported as unchanged, the cached render was re-served, and the stale hunks stayed on screen — including the stale `unmodified lines` counts. Reloading dropped the instance, which is why a refresh always fixed it.

Only snapshots with a grammar were affected: the viewer's plain-text path re-renders unconditionally. HTML, JSON, YAML… hit the cache — exactly the reported case.

## Fix

Pass the screenshot id as the cache key, on both sides of the diff and on the single-snapshot viewer. Side benefit: the worker pool's highlight cache becomes usable, where a single shared key made it both useless and wrong.

## Test

`SwitchingSnapshots` in `DiffEditor.stories.tsx` mounts a `DiffEditor`, switches snapshot in place, and reads the rendered text out of the custom element's shadow DOM.

- with the fix: green
- with the `cacheKey`s removed: red — it times out waiting for the second diff, the first one still showing

Rendering a single snapshot could never have caught this, which is why the story switches in place.

### DIFF
--- a/apps/frontend/src/containers/Build/BuildDiffDetail.tsx
+++ b/apps/frontend/src/containers/Build/BuildDiffDetail.tsx
@@ -792,6 +792,7 @@ function BaseScreenshot({
       }
       return (
         <Snapshot
+          id={diff.baseScreenshot.id}
           url={diff.baseScreenshot.url}
           contentType={diff.baseScreenshot.contentType}
         />
@@ -956,6 +957,7 @@ function CompareScreenshot(props: {
 
       return (
         <Snapshot
+          id={diff.compareScreenshot.id}
           url={diff.compareScreenshot.url}
           contentType={diff.compareScreenshot.contentType}
         />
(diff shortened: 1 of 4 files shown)


gregberge:
May be reverted because you did another fix in #2547

PR #2544, opened 2026-08-24
fix(build): make the ignore flag open its confirmation dialog

jsfez:
Pressing the flag under a build snapshot did nothing, and confirming the dialog left it looking untouched. Two regressions stacked on the same button.

## The dialog never opened

`DialogTrigger` was react-aria's until #2493, where it only provided context to its children — so a controlled one wrapping nothing but a `Modal` was a legitimate shape. The one that replaced it consumes its first element child as the trigger and renders only what follows, so that same shape now renders an empty tree. The modals take `open`/`onOpenChange` directly instead, as every other controlled dialog does since #2540.

Both directions were affected — "Ignore" is only reachable past it because "Don't show this again" skips the dialog.

`DeleteAutomation` was written the same way and went silent with it. `DialogTrigger` now asserts it was given an overlay, so the next one fails loudly instead of rendering nothing — that assertion is what turned up the second call site.

## The flag never flipped

Older, and unrelated to Base UI: #1974 replaced the build page's `useQuery` with `apolloClient.query({ fetchPolicy: "no-cache" })`, so the diff the toolbar holds is not a normalized entity and never hears about the mutation's cache write. The change was ignored on the server while the button stayed as it was until a reload.

`useFragment` reads `ignored` off the cache, falling back to the snapshot until the entity is there — which is also what makes the optimistic response, and its rollback on failure, visible on this page.

The flag is icon-only and had no accessible name; it says which way it goes now.

## Tests

One spec in `tests/project-ignored.spec.ts` walks the whole path on a build page — flag, dialog, confirm, flag flipped, ledger emptied. Each half was reverted in turn to check it fails without it: without the modal fix the dialog heading never appears, without the cache read the flag never flips.

Also driven by hand against a dev server, both directions, no reload.

- `project-ignored`, `project-tests`, `project-automation-pages`, `build`: 40 passed
- `pnpm run static-checks`: 11/11

## Not in here

A diff is only an ignorable change once it carries a fingerprint, which the pipeline computes from the diff image. Seeded rows never go through it, so the flag sits disabled on every snapshot of a freshly seeded dev database and the feature cannot be tried at all. Happy to send that as its own PR.

### DIFF
--- a/apps/frontend/src/containers/Build/toolbar/IgnoreButton.tsx
+++ b/apps/frontend/src/containers/Build/toolbar/IgnoreButton.tsx
@@ -10,7 +10,6 @@ import { graphql } from "@/gql";
 import { useProjectParams } from "@/pages/Project/ProjectParams";
 import { Button, type ButtonProps } from "@/ui/Button";
 import { Checkbox } from "@/ui/Checkbox";
-import { DialogTrigger } from "@/ui/Dialog";
 import {
   Dialog,
   DialogBody,
@@ -157,6 +156,11 @@ function EnabledIgnoreButton(props: {
   const hotkey = useBuildHotkey("ignoreChange", toggle, {
     preventDefault: true,
   });
+  const handleOpenChange = (open: boolean) => {
+    if (!open) {
+      setDialog(null);
+    }
+  };
 
   return (
     <>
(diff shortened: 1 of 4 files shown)


apps/frontend/src/containers/Build/toolbar/IgnoreButton.tsx
```diff
@@ -35,6 +34,21 @@ const IgnoreChangeMutation = graphql(`
...
+ * change would be ignored on the server while the button stayed as it was
+ * until a reload. Both mutations write the change to the cache, so the flag
+ * follows that entity and falls back to the snapshot until it is there.
+ */
+const TestChangeFragment = graphql(`
+  fragment IgnoreButton_TestChange on TestChange {
+    id
+    ignored
+  }
+`);
```
gregberge, line 50: I think we should revert the change to not use the cache to load the page because it could lead to other bugs?

PR #2547, opened 2026-08-25
fix(build): repaint the text diff when the viewer remounts

jsfez:
Follow-up to #2543. Sebastien Lorber reported that selecting another text snapshot *still* leaves the previous diff on screen — now intermittently instead of always. Reproduced on production, on [build 1903](https://app.argos-ci.com/meta-open-source/docusaurus/builds/1903/422242800): the header and the sidebar selection move to the new snapshot, the diff never does, and only a full reload clears it.

#2543 was necessary but not sufficient. Distinct cache keys stopped the renderer from *re-serving* a stale render; they don't make it render at all.

## Cause

Switching snapshot suspends the subtree on the incoming text, and React resumes it onto the **same `<diffs-container>` element**: the old `FileDiff` is torn down (`cleanUp` / `recycle`) and a new one hydrates onto a node that still holds the previous snapshot's markup.

`FileDiff.hydrate` renders only if one of these says so:

```js
shouldRenderCode(pre, hasContent, collapsed)  // !collapsed && pre == null && hasContent
shouldRenderHeader(headerElement, hasContent, disableFileHeader)
```

- `pre` is the leftover `<pre>` from the previous snapshot → first is false.
- We pass `disableFileHeader: true` → second is false.

So `render()` is never called. Instrumenting the live instance shows the end state exactly: the hunks renderer holds the **correct** new diff and no result at all —

```
renderCache = { diff.cacheKey: "390513746:390658128", highlighted: true, result: undefined }
pre == null → false      headerElement == null → true, but disableFileHeader → true
```

— and `render({ forceRender: true })` on that stuck instance repaints the correct snapshot immediately. Without `disableFileHeader` the header path would have forced the render and hidden this entirely, which is why it is specific to us.

## Fix

Key each viewer by the snapshot it shows, so React gives it a fresh element and hydrate finds an empty container. Same trap on the single-snapshot `File`, keyed too.

## Test

`SwitchingSnapshotsThroughSuspense` mirrors the app: the sample text arrives through a Suspense boundary, so switching tears the viewer down and remounts it the way the build page does.

- with the fix: green
- with the keys removed: **red**, times out with the first snapshot still displayed

The existing `SwitchingSnapshots` story keeps passing either way — switching content on a *mounted* viewer repaints normally. That's precisely why it didn't catch this, and why the new one suspends.

### DIFF
--- a/apps/frontend/src/containers/Build/DiffEditor.tsx
+++ b/apps/frontend/src/containers/Build/DiffEditor.tsx
@@ -151,6 +151,15 @@ export function DiffEditor<LAnnotation = undefined>(props: {
   return (
     <Suspense fallback={<SnapshotLoader />}>
       <MultiFileDiff<LAnnotation>
+        // A fresh element per snapshot, and not an optimization to drop. The
+        // viewer hydrates onto whatever DOM its container already holds and
+        // skips its first render when it finds a `<pre>` in there, assuming the
+        // markup is its own; `disableFileHeader` removes the header that would
+        // otherwise force that render. React reuses this element across
+        // snapshots — the subtree suspends on the new text and resumes onto the
+        // same node — so without a key the incoming snapshot inherits the
+        // previous one's markup and never repaints.
+        key={`${originalCacheKey}:${modifiedCacheKey}`}
         oldFile={{
           name: "snapshot",
           contents: original,
@@ -182,6 +191,8 @@ export function Editor(props: {
   return (
     <Suspense fallback={<SnapshotLoader />}>
       <File
+        // See {@link DiffEditor}: same hydrate-onto-stale-DOM trap.
+        key={props.cacheKey}
         file={{
           name: "snapshot",
           contents: props.value,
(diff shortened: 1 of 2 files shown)


apps/frontend/src/containers/Build/DiffEditor.tsx
```diff
@@ -151,6 +151,15 @@ export function DiffEditor<LAnnotation = undefined>(props: {
...
       <MultiFileDiff<LAnnotation>
+        // A fresh element per snapshot, and not an optimization to drop. The
+        // viewer hydrates onto whatever DOM its container already holds and
+        // skips its first render when it finds a `<pre>` in there, assuming the
+        // markup is its own; `disableFileHeader` removes the header that would
+        // otherwise force that render. React reuses this element across
+        // snapshots — the subtree suspends on the new text and resumes onto the
+        // same node — so without a key the incoming snapshot inherits the
+        // previous one's markup and never repaints.
+        key={`${originalCacheKey}:${modifiedCacheKey}`}
```
gregberge, line 162: Then I think your old PR is useless now that we use key?

jsfez: We need both PRs to fix the bug

gregberge: OK

PR #2552, opened 2026-08-25
fix(build): make the ignore flag work again on a build snapshot

jsfez:
Reverts #2544 and redoes it without reading from the Apollo cache, per @gregberge's review there.

Pressing the flag under a build snapshot did nothing, and confirming the dialog left it looking untouched. Two regressions, stacked on the same button.

## The dialog never opened

`DialogTrigger` was react-aria's until #2493, where it only provided context to its children — so a controlled one wrapping nothing but a `Modal` was a legitimate shape. The one that replaced it consumes its first element child as the trigger and renders only what follows, so that same shape now renders an empty tree. The modals take `open`/`onOpenChange` directly instead, as every other controlled dialog does since #2540.

Both directions were affected — "Ignore" is only reachable past it because "Don't show this again" skips the dialog.

`DeleteAutomation` was written the same way and went silent with it. `DialogTrigger` asserts it was given an overlay now, counted on the child slots rather than on the elements that survive filtering, so `{canDelete && <Modal />}` still passes while the old shape throws.

## The flag never flipped

Older, and unrelated to Base UI: #1974 replaced the build page's `useQuery` with `apolloClient.query({ fetchPolicy: "no-cache" })`, so the diff the toolbar holds is not a normalized entity and never hears about the mutation. The change was ignored on the server while the button stayed as it was until a reload.

The button keeps a copy of what the server confirmed — set once the mutation resolves, dropped when the parent's `key` changes with the snapshot. #2544 read it off the cache with `useFragment`; this does the same job without putting the page back on the cache. Worth knowing what that costs: a later fetch of the same change that disagrees stays masked, which only another session can cause, and any future control on this toolbar that writes to a diff will need its own copy. The deep fix is a watched read on the build page, which is what #1974 traded away for paging performance.

## While in there

Both dialogs move onto the shape the ledger's own unignore uses: `role="alertdialog"`, `DialogActionButton` so a pending action blocks dismissal, and a toast either way. A failure keeps the dialog open and says so, where before it closed and said nothing — and since the flag is only set after the server confirms, there is no optimistic state left to roll back wrongly.

The flag is icon-only and had no accessible name. It has a stable one now, with `aria-pressed` carrying the state.

## Tests

Two specs in `tests/project-ignored.spec.ts`: the dialog round trip both ways, and the don't-show-again path that skips the dialog. They locate the flag by its pressed state rather than by copy, so the dialog's confirm buttons cannot be mistaken for it.

Each half of the fix was reverted in turn to check they fail without it: without the modal fix the dialog heading never appears, without the confirmed copy the flag never flips.

- `project-ignored`, `project-tests`, `project-automation-pages`, `build`, `team-settings`: 47 passed
- `pnpm run static-checks`: 11/11

Also driven by hand against a dev server, both directions, no reload.

## Not in here

A diff is only an ignorable change once it carries a fingerprint, which the pipeline computes from the diff image. Seeded rows never go through it, so the flag sits disabled on every snapshot of a freshly seeded dev database and the feature cannot be tried at all. Happy to send that as its own PR.

### DIFF
--- a/apps/frontend/src/containers/Build/toolbar/IgnoreButton.tsx
+++ b/apps/frontend/src/containers/Build/toolbar/IgnoreButton.tsx
@@ -1,5 +1,5 @@
 import { memo, useState } from "react";
-import { useApolloClient, useFragment } from "@apollo/client/react";
+import { useApolloClient } from "@apollo/client/react";
 import { invariant } from "@argos/util/invariant";
 import { FlagOffIcon } from "lucide-react";
 
@@ -10,6 +10,7 @@ import { graphql } from "@/gql";
 import { useProjectParams } from "@/pages/Project/ProjectParams";
 import { Button, type ButtonProps } from "@/ui/Button";
 import { Checkbox } from "@/ui/Checkbox";
+import { DialogTrigger } from "@/ui/Dialog";
 import {
   Dialog,
   DialogBody,
@@ -31,21 +32,6 @@ const IgnoreChangeMutation = graphql(`
       id
       ignored
     }
-  }
-`);
-
-/**
- * What the toolbar reads the flag from. The build page fetches its diffs with
- * `no-cache` (see `useDataState`), so the snapshot it hands down is not a
- * normalized entity and never hears about the mutation's cache write — the
- * change would be ignored on the server while the button stayed as it was
- * until a reload. Both mutations write the change to the cache, so the flag
- * follows that entity and falls back to the snapshot until it is there.
- */
-const TestChangeFragment = graphql(`
-  fragment IgnoreButton_TestChange on TestChange {
-    id
-    ignored
   }
 `);
 
@@ -76,18 +62,11 @@ function EnabledIgnoreButton(props: {
   const params = useProjectParams();
   invariant(params, "IgnoreButton requires project params");
   invariant(diff.change, "IgnoreButton requires a change in the diff");
+  const isIgnored = diff.change.ignored;
   const [dialog, setDialog] = useState<"ignore" | "unignore" | null>(null);
   const auth = useAuth();
   const client = useApolloClient();
   const changeId = diff.change.id;
-  const cachedChange = useFragment({
-    fragment: TestChangeFragment,
-    fragmentName: "IgnoreButton_TestChange",
-    from: { __typename: "TestChange", id: changeId },
-  });
-  const isIgnored = cachedChange.complete
-    ? cachedChange.data.ignored
-    : diff.change.ignored;
 
   const ignoreChange = () => {
     const auditTrailId =
@@ -178,11 +157,6 @@ function EnabledIgnoreButton(props: {
   const hotkey = useBuildHotkey("ignoreChange", toggle, {
     preventDefault: true,
   });
-  const handleOpenChange = (open: boolean) => {
-    if (!open) {
-      setDialog(null);
-    }
-  };
 
   return (
     <>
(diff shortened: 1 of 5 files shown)


apps/frontend/src/containers/Build/toolbar/IgnoreButton.tsx
```diff
@@ -226,14 +188,17 @@ function EnabledIgnoreButton(props: {
...
+            >
               Ignore Change
-            </Button>
+            </DialogActionButton>
           </DialogFooter>
         </Dialog>
       </Modal>
       <Modal open={dialog === "unignore"} onOpenChange={handleOpenChange}>
-        <Dialog size="medium">
+        <Dialog size="medium" role="alertdialog">
```
gregberge: It's not an alert dialog is it?

apps/frontend/src/containers/Build/toolbar/IgnoreButton.tsx
```diff
@@ -191,14 +153,14 @@ function EnabledIgnoreButton(props: {
...
-          aria-label={isIgnored ? "Unignore change" : "Ignore change"}
+          aria-label={hotkey.description}
           aria-pressed={isIgnored}
           onClick={toggle}
           variant={isIgnored ? "danger" : "secondary"}
         />
       </HotkeyTooltip>
       <Modal open={dialog === "ignore"} onOpenChange={handleOpenChange}>
-        <Dialog size="medium">
+        <Dialog size="medium" role="alertdialog">
```
gregberge: It's not an alert dialog is it?

apps/frontend/src/pages/Build/BuildDiffState.tsx
```diff
@@ -526,7 +538,23 @@ function useDataState(props: {
...
   ]);
-  return state;
+
+  // Held beside the diffs, never patched into them: the list memoizes its rows
+  // on diff identity (`memoCreateListItemRow`), so a replaced diff object
+  // leaves the row holding the old one and the snapshot stops looking
+  // selected.
+  const [ignoredChanges, setIgnoredChanges] = useState<Record<string, boolean>>(
+    {},
+  );
```
gregberge: We don't need it, ignored lives on the GraphQL not in a state.

gregberge (changes requested):
We have to revert what has been done in https://github.com/argos-ci/argos/pull/1974 about the cache, not making additional useless complexity here.

apps/frontend/src/pages/Build/BuildDiffState.tsx
jsfez: Agreed it shouldn't live in a state. What I'm unsure about is where to read it from instead, so I'd rather ask than guess again.

`useDataState` fetches the diffs with `fetchPolicy: "no-cache"` and keeps them in React state, so the mutation's cache write never reaches the `diff.change.ignored` this component renders. I checked that dropping just the `no-cache` line does not help either — the query result is still a plain snapshot, so nothing re-renders on the write.

Two ways I can see:

1. `useFragment` on the `TestChange` in `IgnoreButton`, leaving the list as it is. That is what #2544 did before it was reverted.
2. Put `useDataState` back on `useQuery` + `fetchMore` as it was before #1974, which makes the whole list reactive again.

I have 2 working locally, since it matches your review body. One thing worth knowing before it lands: with the list back on the cache, diff objects change identity on every write, and `memoCreateListItemRow` keys its rows without the object — so `activeDiff === item.diff` goes stale and the selected snapshot loses its ring. That is a real bug I hit on this branch, so 2 also means fixing that row cache.

Which of the two do you want?

gregberge: Want we keep as it is but just make sure it's persisted in the Apollo cache? Or we can do the userFragment, it's less bad than an additional state.. <br><br>We should not revert to 2 because it's a big problem for perf.

gregberge (approved):

PR #2557, opened 2026-08-26
fix: repair three modals broken by the Base UI migration

jsfez:
An audit of all 48 `<Modal>` call sites in the frontend, driven by Playwright against the real app: open every reachable modal, check it appears, check it carries exactly one `role="dialog"`, check Escape closes it. Three regressions came out of it, one per commit.

## 1. "Remove from Project" never opened

The same shape as #2540 on the team members list, in the file next door: `<Modal {...removeModal}>` where the object carried `isOpen` and `Modal` takes `open`. The prop arrived `undefined`, `useOverlayRoot` treated the modal as uncontrolled and fell back to `defaultOpen: false`. The menu row fired and nothing happened. A JSX spread skips excess-property checking, so the stray key type-checked clean.

Driven with `useDialogValueState` now, like every other controlled modal.

## 2. The regenerate-token dialog had no accessible name

`RegenerateTokenButton` wrapped `RegenerateTokenDialog` in a `Dialog`, and the component renders one itself — two nested `role="dialog" aria-modal="true"` boxes, the outer one pointing `aria-labelledby` at an id nothing renders. Free before #2493, where the role sat on the popup and a `Dialog` was just a styled box; it is the dialog now.

## 3. A modal opened from a menu lost the Escape that followed

The widest of the three: it hit every modal opened from a menu row — deleting an email, removing a member or a contributor, removing a Discord/Teams channel, deleting an automation.

A row's action opens the dialog and closes the menu in the same tick, but closing only *starts* the popup's exit animation. Traced in the page:

```
t=0ms    the row is pressed
t=8ms    the dialog is mounted
t=185ms  the menu popup is finally unmounted
```

For those ~177ms the menu popup is still there holding focus and the top of the dismiss stack, so an Escape meant for the dialog closes the menu and the dialog stays up. It reproduced 4 times in 8.

Fixed by unmounting the popup outright through the `actionsRef` Base UI's popover root exposes, and only when a row runs an action — the one dismissal with nothing to animate away from, since the pointer has already left. Escape and outside-press keep the animation.

## Verification

Each fix was confirmed by removing it again and watching the check fail. `pnpm test:e2e` is green (132 passed) and `pnpm run static-checks` passes.

The audit spec that found these is not included — it was written to sweep the app, not to be maintained. Worth saying that leaves all three unguarded, #3 especially: it is a race, so it will come back quietly if the menu ever animates out again.

Also noticed while running the suite, unrelated to this branch: under parallel load `personal-settings`, `project-deployments`, `project-ignored` and `project-automation-pages` fail intermittently. They fail the same way on unmodified `main` and pass in isolation.

### DIFF
--- a/apps/frontend/src/containers/Project/Contributors/ProjectContributorsList.tsx
+++ b/apps/frontend/src/containers/Project/Contributors/ProjectContributorsList.tsx
@@ -1,4 +1,3 @@
-import { useState } from "react";
 import { useQuery } from "@apollo/client/react";
 import { invariant } from "@argos/util/invariant";
 
@@ -6,6 +5,7 @@ import { useAssertAuthAccount } from "@/containers/Auth";
 import { ProjectContributorLevelLabel } from "@/containers/ProjectContributor";
 import { RemoveMenu, UserListRow } from "@/containers/UserList";
 import { graphql } from "@/gql";
+import { useDialogValueState } from "@/ui/Dialog";
 import {
   List,
   ListEmpty,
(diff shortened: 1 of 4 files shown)


apps/frontend/src/containers/Project/Contributors/ProjectContributorsList.tsx
gregberge, line 1: Please add E2E tests for these dialogs.

gregberge (approved):

PR #2561, opened 2026-08-28
fix(build): unlock review submission once every reviewable diff is loaded

linhong-dev:
## The Problem
At Cash App, we've recently migrated over to Argos. We have builds that consist of 2,800+ snapshots, and this can be quite annoying to review because:
1. A build can only be approved after all snapshots have loaded
2. Snapshots are loaded in batches of 100

In practice, developers have to open the argos build and wait several minutes for all images to load.

## Proposal
Allow approvals after all reviewable images have loaded. This includes changed/added/removed snapshots.


gregberge (approved):

gregberge:
Reopened in #2563

PR #2562, opened 2026-08-28
fix(build): follow the live ignored state in the sidebar change panel

linhong-dev:
## The Problem
The sidebar's "Change" panel doesn't update when you ignore or unignore a change. It keeps showing the old state until you reload the page. Same bug #2552 fixed for the toolbar flag

## Fix
Same fix as #2552 

| Before | After |
| --- | --- |
| Sidebar still says the change is flaky | Sidebar says the change is ignored |


gregberge (approved):

gregberge:
Reopened in #2564

PR #2588, opened 2026-09-08
chore(dev): fail fast when there is no AWS session

jsfez:
`dev:prod-ro` and `rds-token.sh` both sign an RDS IAM token from your AWS credentials, and without a session the failure only surfaced at connection time — as a password error naming neither AWS nor a fix. Both now check upfront, and the config layer catches a session that expires mid-run.

Also aligns the docs and messages on `aws login`, which is what the team actually uses — `aws sso login` was wrong on current AWS CLI versions.

### DIFF
--- a/apps/backend/src/config/database.ts
+++ b/apps/backend/src/config/database.ts
@@ -28,7 +28,19 @@ function getPassword(config: Config): string | (() => Promise<string>) {
     port: config.get("pg.connection.port"),
     username: config.get("pg.connection.user"),
   });
-  return () => signer.getAuthToken();
+  return async () => {
+    try {
+      return await signer.getAuthToken();
+    } catch (error) {
+      // An AWS session expires while the process keeps running, and the SDK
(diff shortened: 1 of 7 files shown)


apps/backend/src/config/database.ts
```diff
@@ -22,13 +22,26 @@ function getPassword(config: Config): string | (() => Promise<string>) {
     `cannot derive an AWS region from Postgres host "${hostname}" — PG_IAM_AUTH expects an RDS endpoint`,
   );
 
+  const username = config.get("pg.connection.user");
   const signer = new Signer({
     hostname,
     region,
     port: config.get("pg.connection.port"),
-    username: config.get("pg.connection.user"),
```
gregberge, line 29: Why?

gregberge (approved):

apps/backend/src/config/database.ts
jsfez: I revert this change

PR #2590, opened 2026-09-09
feat(analytics): show the Storybook share of screenshots

jsfez:
The Screenshots tile now says how much of the period came from Storybook stories, next to the capture rate — and only for accounts that capture stories, so nobody reads a permanent "0%".

I first tried a dedicated card in the Breakdown section; with only two categories it took more room than it was worth, so the hint carries it instead.

The split also travels in the screenshots CSV export as two columns, and the aggregation exposes it on GraphQL and on `/analytics`.

Left for a follow-up: the Storybook count is still clamped per query because of legacy rows. Fixing the data with a migration and a CHECK constraint would let all three call sites drop the clamp, but that touches the billing table and deserves its own PR.

### DIFF
--- a/apps/backend/src/api/handlers/getAccountAnalytics.e2e.test.ts
+++ b/apps/backend/src/api/handlers/getAccountAnalytics.e2e.test.ts
@@ -139,7 +139,11 @@ describe("getAccountAnalytics", () => {
       .set("Authorization", `Bearer ${patToken}`)
       .expect(200);
 
-    expect(res.body.screenshots.all).toEqual({ total: 0, projects: {} });
+    expect(res.body.screenshots.all).toEqual({
+      total: 0,
+      projects: {},
+      storybook: 0,
+    });
     expect(res.body.screenshots.projects).toEqual([]);
     expect(res.body.builds.all.total).toBe(0);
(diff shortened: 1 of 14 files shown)


gregberge:
> La tuile Screenshots indique la part de la période capturée depuis Storybook, à côté du rythme de capture — et seulement pour les comptes qui ont des stories, sinon la tuile reste celle d'avant.
> 
> J'avais d'abord tenté une carte dédiée dans Breakdown ; à deux catégories elle pesait plus lourd que ce qu'elle disait, le hint suffit.
> 
> Le découpage part aussi dans l'export CSV en deux colonnes, et l'agrégation l'expose sur GraphQL et sur `/analytics`.
> 
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

In english please.

apps/backend/src/api/handlers/getAccountAnalytics.e2e.test.ts
gregberge, line 1: It would be nice to have a E2E test on this. Also the screenshot in Storybook is not very good in terms of size.

gregberge (approved):

PR #2593, opened 2026-09-09
fix(projects): load the latest build without scanning every build

jsfez:
The projects page could take seconds to load for accounts owning a project with a long build history, and the time was spent in a single loader.

## The mechanism

`Project.latestBuild` is backed by a DataLoader that batches project ids, and the query was spelled with `distinct on`:

```sql
select distinct on ("projectId") * from builds
where "projectId" in (…) order by "projectId", "createdAt" desc
```

`distinct on` describes the result, not a way to get it. Postgres implements it as a `Unique` node over a sort, so it reads and sorts _every_ build of _every_ project in the batch, materialising full rows, only to keep the first of each group and discard the rest. On a project with a long history the sort spills to disk and the page waits on it — to render one line per card.

`join lateral … limit 1` expresses the actual intent. Postgres descends the index once per key and stops at the first entry, so the cost follows the number of projects in the batch rather than the number of builds they own.

```sql
select "builds".* from (values (?::bigint), …) as t("projectId")
join lateral (
  select * from "builds" where "builds"."projectId" = t."projectId"
  order by "builds"."createdAt" desc, "builds"."number" desc limit 1
) as builds on true
```

The shape is not new here: three loaders in this same file already use it.

## Scope

Four loaders had the pathological shape — a DataLoader whose only predicate is the very column it groups by:

| loader                                | key                                           |
| ------------------------------------- | --------------------------------------------- |
| `LatestProjectBuild`                  | `projectId` on `builds`                       |
| `LatestChangeDiff`                    | `(testId, fingerprint)` on `screenshot_diffs` |
| `LatestProductionDeploymentByProject` | `projectId` on `deployments`                  |
| `LatestAutomationRun`                 | `automationRuleId` on `automation_runs`       |

The first two read the largest tables in the schema and are the ones that hurt. The last two are on small tables and are moved for the shape, not for a measured win — leaving one behind keeps the trap in the file for the next person to copy.

## What is deliberately not changed

Every other `distinct on` in the backend stays. `distinct on` is only a trap when the grouping column is the _only_ predicate; the remaining sites (`build/siblings.ts`, `git-platform/comment.ts`, `build-notification/aggregated.ts`, `build/strategy/strategies/ci/merge-queue.ts`) are all bounded by a commit or a branch first, so the set they deduplicate is already small. They are correct and fast as they are.

No index is added. `builds` and `screenshot_diffs` already carry exactly the index each lateral needs (`builds_projectid_createdat_number_idx`, `screenshot_diffs_testid_fingerprint_id_desc_notnull_idx`), so those two are a single descent. `deployments` and `automation_runs` have no `createdAt` in their indexes, so their laterals still top-N sort a key's rows — bounded per key instead of across the batch, which is the improvement, but not a descent. Both tables are small enough that a migration would cost more than it buys; the composite index is worth revisiting if either grows.

No shared helper either, though the `(values …) + join lateral` block now appears seven times in the file. The call sites differ in their key arity, casts, filters and echo columns, so a helper would take raw SQL fragments as parameters and hide little. Worth extracting once an eighth appears, but not as part of this change.

## Behaviour

Two orderings gain a tie-breaker and become deterministic: builds on `number desc`, automation runs on `id desc` (deployments already had one). Timestamps are stamped in application code at millisecond precision, so ties are reachable, and `limit 1` makes an untied ordering visibly non-deterministic in a way `distinct on` did not. Both tie-breakers are free — the index already carries the column, or the sort was happening anyway.

## Verification

`loaders.e2e.test.ts` gains cases for `LatestProjectBuild` and `LatestAutomationRun`, which had no coverage and now carry hand-written SQL: newest row per key across a batch, `null` for a key with no rows, and the tie-break. Each was confirmed to fail with the lateral's `order by` inverted, so they are not vacuous. `LatestChangeDiff` was already covered, including the multi-key batching that a lateral rewrite is most likely to break, and `LatestProductionDeploymentByProject` is covered end-to-end by `tests/latestProductionDeployment.e2e.test.ts` through the same GraphQL query the page issues — including a preview-only project, which pins the `production` filter.

The two loaders reading the largest tables now open a Sentry span carrying the batch size, so the effect is observable in production instead of asserted here.

### DIFF
--- a/apps/backend/src/graphql/loaders.e2e.test.ts
+++ b/apps/backend/src/graphql/loaders.e2e.test.ts
@@ -181,6 +181,96 @@ describe("LatestChangeDiff loader", () => {
   });
 });
 
+describe("LatestProjectBuild loader", () => {
+  beforeEach(async () => {
+    await setupDatabase();
+  });
+
+  it("returns the newest build of each project in a batch", async () => {
+    const [projectA, projectB, emptyProject] = await Promise.all([
+      factory.Project.create(),
+      factory.Project.create(),
+      factory.Project.create(),
+    ]);
+    await factory.Build.create({
+      projectId: projectA.id,
+      createdAt: "2026-01-01T00:00:00.000Z",
+    });
+    const newestA = await factory.Build.create({
+      projectId: projectA.id,
+      createdAt: "2026-01-02T00:00:00.000Z",
+    });
+    const newestB = await factory.Build.create({
+      projectId: projectB.id,
+      createdAt: "2025-06-01T00:00:00.000Z",
+    });
+
+    const loaders = createLoaders();
+    const [buildA, buildB, none] = await Promise.all([
+      loaders.LatestProjectBuild.load(projectA.id),
+      loaders.LatestProjectBuild.load(projectB.id),
+      loaders.LatestProjectBuild.load(emptyProject.id),
+    ]);
+
+    expect(buildA?.id).toBe(newestA.id);
+    expect(buildB?.id).toBe(newestB.id);
+    expect(none).toBeNull();
+  });
+
+  it("breaks a tie on `createdAt` with the highest build number", async () => {
+    const project = await factory.Project.create();
+    const createdAt = "2026-01-01T00:00:00.000Z";
+    await factory.Build.create({ projectId: project.id, createdAt });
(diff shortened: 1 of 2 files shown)


gregberge (approved):
If it works, fine to me!

PR #2594, opened 2026-09-09
feat(build): filter on a snapshot variant from its right-click menu

jsfez:
Right-clicking a variant switcher segment adds that value to the filters, and follows it — so the snapshot on screen always matches the filter just set. Where it cannot (walking back out through the switcher, or a filter set from the menu), the sidebar says the snapshot is filtered out rather than leaving the reviewer on a diff no row points at, and an emptied list now names the filters instead of advising the quickstart guide.

Only offers the action for values the filters actually hold — a color scheme every snapshot leaves unset would otherwise crash the chips.

Fixes the sidebar chip menu along the way: a `Menu` is itself a popup, so the one nested inside a `Popover` had no anchor and never left the top-left corner at `opacity: 0`.

CleanShot 2026-09-09 at 17 26 45@2x


apps/frontend/src/pages/Build/metadata/filters/OutOfFilterNotice.tsx
```diff
@@ -0,0 +1,51 @@
...
+ * nothing selected in the list, and the up arrow walking out to the build
+ * overview. Filtering from a variant segment follows the value it names, which
+ * keeps this out of the way for the common case; a filter that empties the
+ * matrix around the snapshot, or one set from the menu above, still lands here.
+ *
+ * Announced rather than corrected: jumping the reviewer somewhere they did not
+ * ask to go loses their place mid-review, which is the one thing worth
+ * protecting.
+ */
+export function OutOfFilterNotice() {
```
gregberge: Where is it displayed?

gregberge (approved):
I think that's a very good feature:

- Some flakies in Argos
- Make sure to add only one screenshot with the menu and a E2E test for filters, but 6 added seems a bit useless?

apps/frontend/src/pages/Build/metadata/filters/OutOfFilterNotice.tsx
jsfez: removed

PR #2595, opened 2026-09-09
feat(billing): alert owners at $200 and $500 of overage without a spend limit

gregberge:
## What

Teams on a usage-based plan that never set a spend limit received no signal about their additional screenshot cost. Owners now get one email when the cost first reaches **$200** and one when it reaches **$500** (in the subscription's currency, like the spend limit itself), inviting them to set up [spend management](https://argos-ci.com/docs/learn/billing-and-subscription/spend-management).

- `claimOverageAlert` (`apps/backend/src/database/services/overage-alert.ts`) picks the highest threshold the period's additional cost has reached and claims it.
- The build job's `updateUsage` calls it after the Stripe usage update and sends the new `overage_alert` notification to the team owners. Accounts with a spend limit are untouched and keep their 50/75/100% alerts.
- New `overage_alert` handler in the configurable `billing` category, whose description is broadened accordingly.
- Migration: `accounts.lastOverageAlertThreshold`, a nullable integer.

## Why it is built this way

- **Once per account, not once per period.** The alert exists to get a limit configured, not to report usage, so it is not repeated every cycle. The existing spend limit alerts use a stateless "cost crossed the threshold since the last Stripe update" comparison, which cannot express "once ever", hence the stored state.
- **A single integer column** holding the highest threshold already sent. A build jumping past both thresholds sends only the $500 email. Rejected: querying `notification_workflows` (a JSON log keyed by slugs that can change, not state) and a dedicated table (more machinery than two thresholds warrant).
- **Claimed with a conditional `UPDATE … RETURNING`.** The job loads the account before taking its per-account Redis lock, so the in-memory value can be stale. The update makes the check itself, and no row returned means another run already claimed it. Skipping early on the loaded value is still safe because a claim only ever raises it.
- **Claimed after the Stripe update.** A claim survives a failed run, so it comes after the step that can fail and get retried.
- Thresholds use `>=` and are formatted in the subscription currency ($200 / €200), mirroring how `meteredSpendLimitByPeriod` is interpreted.

## Verification

- `overage-alert.e2e.test.ts` covers: below threshold, single claim, growing cost, both thresholds at once, spend limit set, two concurrent claims, flat-rate plan.
- `pnpm run static-checks` passes.
- Both variants rendered through `/notification-preview/overage_alert`; subject, settings link (`/<slug>/settings#spend-management`) and docs link checked.

## Screenshots

Copy revised after review to read as a heads-up about the coming invoice rather than a warning. It speaks of overage rather than of what is metered, so billing something new does not date it, and the key facts are in bold.

$200 variant:

[](https://app.argos-ci.com/m/ioffv1uwufadi59gigq2)

$500 variant:

[](https://app.argos-ci.com/m/2kmx5yixfuwq7ixmt7hs)

## Not in this PR

- The spend management docs page does not mention these alerts yet.
- After pulling, run `pnpm run --filter @argos/backend db:migrate:latest` (and the `NODE_ENV=test` variant for the Playwright database).

### DIFF
--- a/apps/backend/db/migrations/20260909153631_overage-alert.js
+++ b/apps/backend/db/migrations/20260909153631_overage-alert.js
@@ -1,0 +1,17 @@
+/**
+ * @param {import('knex').Knex} knex
+ */
+export const up = async (knex) => {
+  await knex.schema.alterTable("accounts", (table) => {
+    table.integer("lastOverageAlertThreshold");
+  });
+};
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+export const down = async (knex) => {
+  await knex.schema.alterTable("accounts", (table) => {
+    table.dropColumn("lastOverageAlertThreshold");
+  });
+};
--- a/apps/backend/db/structure.sql
+++ b/apps/backend/db/structure.sql
@@ -2,7 +2,7 @@
 -- PostgreSQL database dump
 --
 
-\restrict KpjS2jKXTxQgaV2MMwS992ZcqvSv3J5pYvxbsy4hL6CE5lxMr5e9WGe0MUzJrDr
+\restrict 9RtOxyeUu8jhzkjsBQQ5Ur9nRozyDFBofOxHaaPMoGifWmN1aiXM3dnH2fEDWFd
 
 -- Dumped from database version 18.4
 -- Dumped by pg_dump version 18.4 (Homebrew)
@@ -127,6 +127,7 @@ CREATE TABLE public.accounts (
     "meteredSpendLimitByPeriod" integer,
     "blockWhenSpendLimitIsReached" boolean DEFAULT false NOT NULL,
     "originInstallationId" bigint,
+    "lastOverageAlertThreshold" integer,
     CONSTRAINT accounts_only_one_owner CHECK ((num_nonnulls("userId", "teamId") = 1))
 );
 
@@ -6692,7 +6693,7 @@ ALTER TABLE ONLY public.users
 -- PostgreSQL database dump complete
 --
 
-\unrestrict KpjS2jKXTxQgaV2MMwS992ZcqvSv3J5pYvxbsy4hL6CE5lxMr5e9WGe0MUzJrDr
+\unrestrict 9RtOxyeUu8jhzkjsBQQ5Ur9nRozyDFBofOxHaaPMoGifWmN1aiXM3dnH2fEDWFd
 
 -- Knex migrations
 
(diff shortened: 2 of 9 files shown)


jsfez:
## Email copy: inform rather than warn

The goal of these emails is to give owners a heads-up so the next invoice holds no surprise. The current copy reads more like a warning about exceeding a limit, and almost as advice not to go over. A few phrases give that impression:

- "has used more screenshots than its plan includes" frames the overage as a fault, while it is how a usage-based plan is meant to work.
- "nothing caps this cost" and "so your invoice never goes above what you expect" suggest the overage is something to avoid.
- "this is the last alert Argos sends without one" reads as an ultimatum.
- "has reached" in the subject borrows the spend limit vocabulary, where a threshold the customer set was crossed. Here it is an information point.

The existing 50% spend limit email already has the right tone ("a friendly reminder to help you monitor your usage and avoid any surprises"). The proposal below is aligned on it.

### $200 variant

**Subject:** `Heads-up: $200 of additional screenshots this billing period`
**Preview:** `A quick update on {accountName}'s usage, so your next invoice holds no surprise.`
**Title:** `$200 of additional screenshots this billing period`

> Hi {name},
>
> Your team, **{accountName}**, has used all the screenshots included in its plan this billing period, plus additional screenshots for a total of **$200** so far. They are billed at your plan's per-screenshot rate and will appear on your next invoice.
>
> This is exactly what your usage-based plan is for: your builds keep running and you only pay for what your team actually uses. We simply want to make sure nothing on your invoice comes as a surprise.
>
> If you would like more visibility on this, you can set a spend limit from your team settings. Argos will then let you know when you reach 50%, 75% and 100% of the amount you choose, and can pause builds at that point if you prefer.
>
> [Set up spend management]
>
> You can check your current usage at any time in your [team settings], and learn more about [spend management] in the documentation. If you have any questions about your usage or your plan, don't hesitate to reach out.

### $500 variant

Shorter: the $200 email carries the explanation, no need to repeat it. It still has to stand on its own, because the two emails can be months apart (a threshold is claimed once per account, not per period), a single build can jump past both thresholds and send only this one, and an owner added in between never got the first.

**Subject:** `Heads-up: $500 of additional screenshots this billing period`
**Title:** `$500 of additional screenshots this billing period`

> Hi {name},
>
> Your team, **{accountName}**, has now used **$500** of additional screenshots this billing period, on top of the ones included in its plan. They are billed at your plan's per-screenshot rate on your next invoice, and your builds keep running.
>
> Argos sends this heads-up at $200 and $500 of additional usage, and this is the second and last one. If you would like to keep being notified as your usage grows, you can set a spend limit: Argos will then let you know at 50%, 75% and 100% of the amount you choose, and can pause builds at that point if you prefer.
>
> [Set up spend management]
>
> If you have any questions about your usage or your plan, don't hesitate to reach out.

"Argos sends this heads-up at $200 and $500" works whether the recipient got the first email yesterday, months ago, or never.

### Optional

The payload only carries the amount. Adding the number of additional screenshots and the per-screenshot rate would make the email more concrete. Not needed to fix the tone.

jsfez (changes requested):
See my comment

jsfez (approved):

PR #2596, opened 2026-09-09
fix(build): drop the story mode switcher when it restates the color scheme

jsfez:
A Storybook mode is commonly named after the scheme it sets, and the SDK records both — so the toolbar carried two switchers and four buttons for one binary choice, and both of them landed on the same snapshots.

Judged on the names rather than on how the two axes cut the snapshots up: a `mobile`/`desktop` pair can line up one-to-one with `light`/`dark` by accident of the matrix, and dropping it on that evidence would take away the labels that tell the reviewer what they are looking at. The mode stays stated in the Metadata panel either way.

### DIFF
--- a/apps/frontend/src/pages/Build/metadata/utils.test.ts
+++ b/apps/frontend/src/pages/Build/metadata/utils.test.ts
@@ -8,6 +8,7 @@ import {
   getUniqueColorSchemes,
   getUniqueStoryModes,
   getUniqueViewports,
+  storyModeRestatesColorScheme,
 } from "./utils";
 
 /**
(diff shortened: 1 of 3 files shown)


jsfez:
Closing: the rule infers what a mode contains from what it is named, and a mode name is a free label over an arbitrary bundle of globals — Chromatic's own example is `'dark desktop': { backgrounds, theme, viewport }`, and baselines are keyed on the name alone. A mode named `dark` can perfectly well be `{ theme: 'dark', locale: 'ko' }`, and dropping the switcher there tells the reviewer they are changing the color scheme while they are also changing the locale — and takes away the name that identifies the variant.

It also never fires on the case it is named after: `@argos-ci/storybook` does not report `colorScheme` at all, so it resolves to `light` for every snapshot, a `dark` mode never matches, and the color scheme switcher does not render in the first place. The duplication needs an SDK that reports both.

If it is worth fixing, the fix is upstream: have the SDK say what the mode set, so Argos knows instead of guessing.

PR #2606, opened 2026-09-14
chore(ui): remove Text in favour of named layout components

gregberge:
Two commits. The first is the dead-code removal this PR started as; the second is the simplification it uncovered, added after review.

## 1. Drop the inert `slot` DOM attribute (`3eedb1c57`)

`Text` rendered `slot={slot}` on its `<span>` to feed two Tailwind selectors in `ListBox.tsx`:

```
has-[[slot=description]]:flex-wrap **:[[slot=label]]:truncate
```

Both died with react-aria in 45f1c2df5 (`feat(ui): move Select and its list to Base UI`). Nothing replaced them, so the attribute reached the DOM and did nothing there — a light-DOM `slot` is inert without a shadow root. The comment defending it outlived what it described, and was already half-wrong before the migration: it cites `slot=label`, but `TextSlot` is `"headline" | "description"`.

Verified nothing read it: no `[slot=` / `[slot]` / `group-has-[slot` anywhere, no `slot` in any `.css`, no `getAttribute("slot")` / `assignedSlot` / `::slotted`, no occurrence in `tests/`, no story or snapshot, and Base UI has no notion of the prop the way react-aria's `useSlottedContext` did.

## 2. Remove `Text` entirely (`cdcba874c`)

Prompted by review. `slot` was never ignored — it was the key into the layout's `slots` map, and `useTextStyle` read `context.slots[slot]`. But checking that turned up the real problem:

- **Every provider declares exactly one slot.** `PageHeader` offers only `headline`, `EmptyState` only `description`. The map is built for many-per-layout and never holds more than one entry.
- **Every call site sits lexically inside the layout that styles it.** `<EmptyState>` directly wraps its own `<Text slot="description">`. Nothing adapts across layouts — the one thing a context buys over a named component.

So the styles move to components that say what they are, next to the layouts that own them:

| was | now | classes |
| --- | --- | --- |
| `<Text slot="headline">` ×16 | `PageHeaderHeadline` | `text-low text-sm` |
| `<Text slot="description">` ×27 | `EmptyStateDescription` | `text-low max-w-lg text-center text-sm text-balance` |
| `<Text>` ×5 | `InviteDescription` | `text-low` |

`TextContext`, `TextSlot` and `useTextStyle` go with it, along with the runtime invariant for a slot the surrounding layout didn't declare — unreachable once there is no name to get wrong.

## Render-identical

Each component carries the exact classes its context supplied, all three render a `span`, and a context provider emits no DOM of its own. The 5 slot-less call sites were checked individually: all sit inside `InviteContainer`, so all were already receiving `text-low`. **Argos should report zero visual diffs** — that's the assertion to check this PR against.

## Notes for reviewers

`chore`, not `fix`: nothing a user can observe changes in either commit.

`pnpm run static-checks` passes 11/11. Worth knowing in a fresh worktree: it is not self-contained — knip loads `apps/backend/knexfile.js`, which imports the **built** `dist/config`, so without a backend build it throws `cannot find configuration param 'target'` and bails out into a bogus "253 unused files" + unused-`pg` report. Run `pnpm turbo run build --filter=@argos/backend` first.


apps/frontend/src/ui/Text.tsx
```diff
@@ -21,13 +21,7 @@ type TextContextValue = TextStyle & {
...
-/**
- * Text whose style comes from the layout it sits in.
- *
- * `slot` stays on the element: `ListBoxItem` selects on it
- * (`has-[[slot=description]]:flex-wrap`, `**:[[slot=label]]:truncate`), so
- * dropping it from the DOM would break that styling silently.
- */
+/** Text whose style comes from the layout it sits in. */
 export function Text({
   slot,
```
gregberge: It seems we are ignoring slot now, what is the point of having it? Can we completely get rid of Text?

gregberge: One correction first, because it changes the reasoning: `slot` wasn't being ignored. It was the key into the layout's `slots` map — `useTextStyle` read `context.slots[slot]`, and the styles came from `Layout.tsx:13` (`headline`) and `:50` (`description`). The attribute this PR removed was only ever a *copy* of that key on the DOM, read by two Tailwind selectors that died with react-aria. Styling never went through it.

But your instinct was right, for a different reason than the one in the comment. Two things I found while checking:

- **Every provider declares exactly one slot.** `PageHeader` offers only `headline`, `EmptyState` only `description`. The map is built for many-per-layout and never holds more than one entry.
- **Every call site sits lexically inside the layout that styles it** — `<EmptyState>` directly wraps its own `<Text slot="description">`. Nothing adapts across layouts, which is the one thing a context buys over a named component.

So `<Text slot="description">` was a long way of writing "the description of the `EmptyState` I'm already inside". `Text` is gone now, replaced by components that say what they are, next to the layouts that own the styles:

| was | now | classes |
| --- | --- | --- |
| `<Text slot="headline">` ×16 | `PageHeaderHeadline` | `text-low text-sm` |
| `<Text slot="description">` ×27 | `EmptyStateDescription` | `text-low max-w-lg text-center text-sm text-balance` |
| `<Text>` ×5 | `InviteDescription` | `text-low` |

`TextContext`, `TextSlot` and `useTextStyle` go with it — including the runtime invariant for a slot the layout didn't declare, which stops being reachable once there's no name to get wrong.

Render-identical by construction: each component carries the exact classes its context supplied, all three render a `span`, and a context provider emits no DOM of its own. I checked the 5 slot-less ones individually — all were inside `InviteContainer`, so all were already getting `text-low`. Argos should show zero diffs; if it doesn't, I've missed something and want to know.

